PR #32, opened 2020-05-18
Add provider and namespace to stack

gergu:


gergu:
[Add required fields (provider + namespace) to TF provider](https://trello.com/c/A93MHXf6/26-add-required-fields-provider-namespace-to-tf-provider)

spacelift/resource_stack.go
```diff
@@ -76,16 +76,26 @@ func resourceStack() *schema.Resource {
...
-				Description: "Name of the GitHub repository, without the owner part",
+				Description: "Name of the repository, without the owner part",
 				Required:    true,
 			},
 			"terraform_version": &schema.Schema{
 				Type:        schema.TypeString,
 				Description: "Terraform version to use",
 				Optional:    true,
 			},
+			"vcs_provider": &schema.Schema{
```
marcinwyszynski, line 94: Can we make GitHub a default here.

spacelift/resource_stack.go
```diff
@@ -76,16 +76,26 @@ func resourceStack() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Description: "VCS provider of the repository",
+				Required:    true,
```
marcinwyszynski, line 97: If we add a default, there's no need to mark it as `required`.

spacelift/resource_stack.go
```diff
@@ -76,16 +76,26 @@ func resourceStack() *schema.Resource {
 				Required:    true,
 				ForceNew:    true,
 			},
+			"namespace": &schema.Schema{
```
marcinwyszynski, line 79: This should only be required for Gitlab.

marcinwyszynski:
In principle, there should be a way of making this change a no-op for existing definitions.

spacelift/resource_stack.go
gergu: Is there any way to verify that on a schema definition level? As far as I see the value of  `Required` and `Optional` can be `bool` type only. 

I'm going to achieve the goal by setting `Optional: true` for the `namespace` and then verifying that in `resourceStackCreate` and `resourceStackUpdate` functions but if there's better way to do it, please let me know.

marcinwyszynski: Unfortunately that's a known limitation of Terraform for now - [you can only validate individual fields](https://godoc.org/github.com/hashicorp/terraform/helper/schema#SchemaValidateFunc). So the "standard" way of doing it would indeed be on `create` and `update` actions, which sucks because you don't get any advance warning from the plan.

I can see two solutions if you want to provide a better user experience.

1) Combine provider and namespace, and make it a single field. Perhaps you can still call it a "namespace" but expect an input formatted like `GITLAB/spacelift`. Pretty ugly, right?

2) Make `gitlab` itself a field. So a stack definition for Gitlab would look like:

```hcl
resource "spacelift_stack" "gitlab_stack" {
  gitlab {
    namespace = "spacelift"
  }
  
  repository = "infra"

  # other fields...
}
```

This would require something like that on the provider end:

```go
"gitlab": &schema.Schema{
	Type:     schema.TypeList,
	Optional: true,
	MaxItems: 1,
	Elem: &schema.Resource{
		Schema: map[string]*schema.Schema{
			"namespace": {
				Type:     schema.TypeString,
				Required: true,
			},
		},
	},
},
```

...and there would be some logic involved in building the right GraphQL query from that input. This is a *much* nicer solution, though it does require a bit more work.

spacelift/data_stack.go
```diff
@@ -67,6 +80,11 @@ func dataStack() *schema.Resource {
 				Description: "Terraform version to use",
 				Computed:    true,
 			},
+			"vcs_provider": &schema.Schema{
```
marcinwyszynski, line 83: You don't need this field now. If the `gitlab` list is empty, it implies that VCS provider is GitHub.

PR #34, opened 2020-07-08
Fix gitlab deletion ghost in github stacks

cube2222:


cube2222:
[Why `gitlab` block does not get deleted?](https://trello.com/c/vKPgEIW7/118-why-gitlab-block-does-not-get-deleted)

PR #36, opened 2020-08-03
Add worker pool data and resource

b4k3r:


README.md
```diff
@@ -709,4 +710,61 @@ In addition to all arguments above, the following attributes are exported:
...
+
+- `name` - (Required) - The name of the worker pool;
+- `description` - (Optional) - The description of the worker pool;
+
+#### Attributes reference
+
+In addition to all arguments above, the following attributes are exported:
+
+- `id` - The immutable ID (slug) of the worker pool;
+- `config` - The credentials necessary to connect WorkerPool's workers to the control plane (this config is visible only once after resource creation);
```
marcinwyszynski, line 768: Attributes are generally only visible after resource creation, that's the whole point. So no need to add this note here.

spacelift/resource_worker_pool.go
```diff
@@ -0,0 +1,129 @@
...
+			"name": &schema.Schema{
+				Type:        schema.TypeString,
+				Description: "name of the worker pool",
+				Required:    true,
+			},
+			"description": &schema.Schema{
+				Type:        schema.TypeString,
+				Description: "description of the worker pool",
+				Optional:    true,
+				Default:     "",
```
marcinwyszynski, line 38: I don't think you actually want to set a default value to an empty string. Instead, I'd suggest checking for the field value being set.

PR #76, opened 2020-11-03
Allow users to rename a stack

marcinwyszynski:


cube2222 (approved):
If it works 🤷🏼‍♂️

marcinwyszynski:
Yes that will work, it just doesn't change the immutable slug.

PR #81, opened 2020-11-12
More universal, shorter policy types

marcinwyszynski:


spacelift/resource_policy.go
```diff
@@ -11,6 +11,25 @@ import (
 	"github.com/spacelift-io/terraform-provider-spacelift/spacelift/internal/structs"
 )
 
+var policyTypes = []string{
```
cube2222, line 14: Please add a comment about what's happening here. Why there are two access policies, why there's the type name replacements, and how they interact.

It took me a while of thinking to understand what's happening here.

cube2222 (approved):

spacelift/resource_policy.go
marcinwyszynski: Done.

PR #90, opened 2020-11-18
Add before_init and runner_image runtime fields to the Stack

b4k3r:


marcinwyszynski (approved):

spacelift/data_stack.go
```diff
@@ -82,6 +88,11 @@ func dataStack() *schema.Resource {
 				Description: "Name of the repository, without the owner part",
 				Computed:    true,
 			},
+			"runner_image": {
+				Type:        schema.TypeString,
+				Description: "Name of the Docker image used to process the Run",
```
cube2222, line 93: "used to process Runs" there's no single Run.

spacelift/data_stack.go
```diff
@@ -156,6 +173,12 @@ func dataStackRead(d *schema.ResourceData, meta interface{}) error {
 		d.Set("project_root", nil)
 	}
 
+	if stack.RunnerImage != nil {
+		d.Set("runner_image", *stack.RunnerImage)
```
cube2222, line 177: I'm obviously not sure, but wouldn't just d.Set("runner_image", stack.RunnerImage) without the if around it work?

cube2222 (approved):
With nits.

spacelift/data_stack.go
b4k3r: You are right, I've applied it to other fields as well.

PR #100, opened 2020-12-07
Add worker pools data source

michieldewilde:


spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,78 @@
...
+		return errors.Wrap(err, "could not query for worker pools")
+	}
+
+	workerPools := query.WorkerPools
+	if workerPools == nil {
+		return errors.New("worker pools not found")
+	}
+
+	wps := flattenDataWorkerPoolsList(workerPools)
+	if err := d.Set("worker_pools", wps); err != nil {
```
michieldewilde, line 59: @marcinwyszynski this sets the `worker_pools` Attribute on the schema.
Yet the it doesn't have the attribute while testing.

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,78 @@
...
+		return errors.Wrap(err, "could not set worker pools")
+	}
+
+	return nil
+}
```
marcinwyszynski, line 64: You'll still need to set `SetID` (eg. set it to the current timestamp), otherwise the resource will not be persisted. [Here's](https://github.com/spacelift-io/terraform-provider-spacelift/blob/master/spacelift/data_ips.go#L25) an example.

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,78 @@
...
+					Schema: map[string]*schema.Schema{
+						"worker_pool_id": {
+							Type:        schema.TypeString,
+							Description: "ID of the worker pool",
+							Required:    true,
+						},
+						"config": {
+							Type:        schema.TypeString,
+							Description: "credentials necessary to connect WorkerPool's workers to the control plane",
+							Computed:    true,
```
marcinwyszynski, line 28: I don't think you need to mark anything as computed except `worker_pools`.

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,78 @@
...
+							Sensitive:   true,
+						},
+						"name": {
+							Type:        schema.TypeString,
+							Description: "name of the worker pool",
+							Computed:    true,
+						},
```
marcinwyszynski, line 35: What about `description`?

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,78 @@
...
+	wps := flattenDataWorkerPoolsList(workerPools)
+	if err := d.Set("worker_pools", wps); err != nil {
+		return errors.Wrap(err, "could not set worker pools")
+	}
+
+	return nil
+}
+
+func flattenDataWorkerPoolsList(workerPools []*structs.WorkerPool) []map[string]interface{} {
+	wps := make([]map[string]interface{}, 0, len(workerPools))
```
marcinwyszynski, line 67: Nit: if the length of the resulting slice is known, why not preallocate it directly, and do `wps[index]` instead of of `append`'ing?

spacelift/provider.go
```diff
@@ -62,6 +62,7 @@ func Provider() *schema.Provider {
 			"spacelift_stack_aws_role":            dataStackAWSRole(),           // deprecated
 			"spacelift_stack_gcp_service_account": dataStackGCPServiceAccount(), // deprecated
 			"spacelift_worker_pool":               dataWorkerPool(),
+			"spacelift_worker_pools":               dataWorkerPools(),
```
marcinwyszynski, line 65: Formatting (weird alignment).

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,94 @@
...
+		Schema: map[string]*schema.Schema{
+			"worker_pools": &schema.Schema{
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"worker_pool_id": {
+							Type:        schema.TypeString,
+							Description: "ID of the worker pool",
+							Required:    true,
```
cube2222, line 23: If it's a data source, then parameters which are read and don't have to be provided, should be marked as Computer: true, instead of Required: true

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,94 @@
...
+	d.SetId("spacelift-worker-pools")
+
+	return nil
+}
+
+func flattenDataWorkerPoolsList(workerPools []*structs.WorkerPool) []map[string]interface{} {
+	wps := make([]map[string]interface{}, len(workerPools))
+
+	for index, wp := range workerPools {
+		var description interface{}
```
marcinwyszynski, line 77: Nit: I think it can be a string pointer?

marcinwyszynski (approved):
@cube2222 has a point re: `computed`, otherwise looks reasonable.

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,94 @@
...
+		WorkerPools []*structs.WorkerPool `graphql:"workerPools()"`
+	}
+	variables := map[string]interface{}{}
+
+	if err := meta.(*internal.Client).Query(&query, variables); err != nil {
+		return errors.Wrap(err, "could not query for worker pools")
+	}
+
+	workerPools := query.WorkerPools
+	if workerPools == nil {
```
cube2222, line 59: Quite sure this check is unnecessary. I suppose it's ok if the list is empty?

cube2222 (changes requested):
Looks good, other than the two comments

spacelift/data_worker_pools.go
michieldewilde: I would say that returning early would be better way to handle the scenario when there are no worker pools?

marcinwyszynski: Returning early should be OK, but returning an error isn't - having no worker pools is a perfectly valid (though weird) scenario, so no error necessary here.

marcinwyszynski (approved):
Merge away, let's try to use it in anger.

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil
```
marcinwyszynski, line 60: Actually, I think you will still want to `SetID` here.

cube2222: You sure? I think you'd want to `d.Set("worker_pools", nil)`

cube2222: The list of workerpools exists, and is non-null. But the list is empty.

cube2222: @marcinwyszynski ^

cube2222 (approved):

spacelift/data_worker_pools.go
marcinwyszynski: True that.

marcinwyszynski: But the ID should still be there IMO?

cube2222: Yea, it's set before, so that's alright.

cube2222: It *was* set before, yea, it has to be set there.

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	wps := flattenDataWorkerPoolsList(workerPools)
+	if err := d.Set("worker_pools", wps); err != nil {
```
michieldewilde, line 66: Do we want to set `worker_pools` to nil here as well? @cube2222 @marcinwyszynski

cube2222: Why so?

cube2222: You mean if the error is non-nil?

cube2222: @michieldewilde

marcinwyszynski: It's just set ID to "", which will remove the entire resource.

spacelift/data_worker_pools.go
```diff
@@ -0,0 +1,95 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataWorkerPoolsRead(d *schema.ResourceData, meta interface{}) error {
+	d.SetId("spacelift-worker-pools")
```
marcinwyszynski, line 49: I'd move it after the query.

spacelift/data_worker_pools.go
michieldewilde: My idea is that there are two scenarios:
- no worker_pools found => set worker_pools resource to nil and keep the resource ID
- generic error => set resource ID to `""` and return error

spacelift/data_worker_pools.go
marcinwyszynski: So that it's only set if the query succeeds.

michieldewilde:
@cube2222  @marcinwyszynski should be fine now.

cube2222 (approved):
LGTM

marcinwyszynski (approved):

PR #101, opened 2020-12-08
Handle Terraform 0.14 new plugin location.

cube2222:


b4k3r (approved):

marcinwyszynski (changes requested):
No need. What's more, we want to remove the previous line, too, which was only necessary because we didn't have a public registry. Now that we do have it, this is useless.

PR #172, opened 2021-05-13
Add showcases support.

cube2222:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]()

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/spacelift/data_stack.go
+++ b/spacelift/data_stack.go
@@ -154,7 +154,7 @@ func dataStack() *schema.Resource {
 				Description: "Name of the Docker image used to process Runs",
 				Computed:    true,
 			},
-			"showcases": {
+			"showcase": {
 				Type:        schema.TypeList,
 				Description: "Showcase-related attributes",
 				Computed:    true,
@@ -234,13 +234,13 @@ func dataStackRead(ctx context.Context, d *schema.ResourceData, meta interface{}
 			return diag.Errorf("error setting gitlab (resource): %v", err)
 		}
 	}
-	if stack.Provider == "SHOWCASES" {
+	if stack.Provider == "SHOWCASE" {
 		m := map[string]interface{}{
 			"namespace": stack.Namespace,
 		}
 
-		if err := d.Set("showcases", []interface{}{m}); err != nil {
-			return diag.Errorf("error setting showcases (resource): %v", err)
+		if err := d.Set("showcase", []interface{}{m}); err != nil {
+			return diag.Errorf("error setting showcase (resource): %v", err)
 		}
 	}
 
--- a/spacelift/data_stack.go
+++ b/spacelift/data_stack.go
@@ -154,6 +154,20 @@ func dataStack() *schema.Resource {
 				Description: "Name of the Docker image used to process Runs",
 				Computed:    true,
 			},
+			"showcases": {
+				Type:        schema.TypeList,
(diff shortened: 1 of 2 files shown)


spacelift/data_stack.go
```diff
@@ -154,6 +154,20 @@ func dataStack() *schema.Resource {
 				Description: "Name of the Docker image used to process Runs",
 				Computed:    true,
 			},
+			"showcases": {
+				Type:        schema.TypeList,
+				Description: "Showcase-related attributes",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"namespace": {
```
marcinwyszynski, line 163: What does `namespace` stand here for?

spacelift/data_stack.go
```diff
@@ -154,6 +154,20 @@ func dataStack() *schema.Resource {
 				Description: "Name of the Docker image used to process Runs",
 				Computed:    true,
 			},
+			"showcases": {
```
marcinwyszynski, line 157: Why plural?

cube2222: Because we may have multiple showcases in the future? But you're right, maybe it should still read Showcase.

spacelift/data_stack.go
cube2222: Now it will always be spacelift-io. In the future we might want to have showcases in a different org, so I left this be.

marcinwyszynski: Wouldn't it be like an ID of a showcase? Like, you could have a showcase with a Minecraft server or something similar?

spacelift/data_stack.go
marcinwyszynski: Still, that's just a block name.

spacelift/data_stack.go
cube2222: Yeah, but I assume showcases are stored in GitHub, and we just put namespace+repo names into an allow list.

spacelift/resource_stack.go
```diff
@@ -383,15 +407,22 @@ func stackInput(d *schema.ResourceData) structs.StackInput {
...
 	}
-	if !foundGitlab {
+	if showcase, ok := d.Get("showcase").([]interface{}); ok {
+		if len(showcase) > 0 {
+			foundSomethingElseThanGitHub = true
+			ret.Namespace = toOptionalString(showcase[0].(map[string]interface{})["namespace"])
+			ret.Provider = graphql.NewString(vcsProviderShowcases)
+		}
+	}
+	if !foundSomethingElseThanGitHub {
```
adamconnelly, line 425: Just a thought - could you default `ret.Provider` to `GITHUB` before doing the logic for the other VCS providers? That way you could get rid of this variable, and it means when someone is adding another provider block they don't need to remember to set `foundSomethingElseThanGitHub`.

adamconnelly (approved):
Looks good - just got a really minor code style nit / suggestion.

spacelift/resource_stack.go
cube2222: Good point, done.

marcinwyszynski (approved):

spacelift/data_stack.go
marcinwyszynski: I get it now, thanks.

PR #178, opened 2021-05-25
Run resource

marcinwyszynski:
## Description of the change

This PR adds the ability to trigger a tracked run declaratively.

## Type of change

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/spacelift/provider.go
+++ b/spacelift/provider.go
@@ -77,6 +77,7 @@ func Provider(commit, version string) plugin.ProviderFunc {
 				"spacelift_mounted_file":              resourceMountedFile(),
 				"spacelift_policy_attachment":         resourcePolicyAttachment(),
 				"spacelift_policy":                    resourcePolicy(),
+				"spacelift_run":                       resourceRun(),
 				"spacelift_stack":                     resourceStack(),
 				"spacelift_stack_destructor":          resourceStackDestructor(),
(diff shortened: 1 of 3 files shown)


adamconnelly (approved):
Looks good.

Really minor question / suggestion: is it worth (and are we even able to) do a test to make sure the resource isn't recreated unless the keepers change? Maybe not mega worthwhile since it's just terraform functionality, but figured it was worth asking.

cube2222 (approved):

cube2222:
> Really minor question / suggestion: is it worth (and are we even able to) do a test to make sure the resource isn't recreated unless the keepers change? Maybe not mega worthwhile since it's just terraform functionality, but figured it was worth asking.

That would be testing terraform though, wouldn't it? We've marked the ForceNew in the correct places, so it will be recreated only if one of those parameters changes (so not only keepers, just whichever attribute)

b4k3r (approved):

adamconnelly:
> That would be testing terraform though, wouldn't it?

Yeah - that's why I was on the fence about it. I think you could view it both ways: either it would be testing Terraform's functionality, which is pointless; or it would be testing that we're using Terraform correctly (i.e. we've marked the field in such a way that it forces a recreate).

I'm not going to be upset either way 😄

PR #291, opened 2022-05-16
Validate module name.

cube2222:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]()

### Development

- [x] Lint rules pass locally

### Code review

- [ ] The target branch is `future` unless the change is going directly into production


spacelift/resource_module_test.go
```diff
@@ -114,4 +115,30 @@ func TestModuleResource(t *testing.T) {
...
 		})
 	})
+
+	t.Run("invalid name", func(t *testing.T) {
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+
+		config := func(description string, protectFromDeletion bool) string {
+			return fmt.Sprintf(`
+				resource "spacelift_module" "test" {
+					name                  = "github-Module-%s"
```
adamconnelly, line 125: I'd be tempted to remove all the optional properties from this (e.g. `administrative`, `description`, etc) to make it more clear what we're testing (i.e. the name), and since you've got a helper function maybe add a few more test cases (e.g. empty name, a name ending in a `-`, etc).

I'm guessing we don't need to add the random ID as a suffix for the invalid name tests because they shouldn't even get as far as making an API call, right?

spacelift/resource_module_test.go
```diff
@@ -114,4 +115,30 @@ func TestModuleResource(t *testing.T) {
...
 	})
+
+	for _, name := range []string{
+		"github-Module",
+		"github-module-",
+		"_github-module",
+		"0github-module",
+	} {
+		t.Run("invalid name", func(t *testing.T) {
+			config := func(description string, protectFromDeletion bool) string {
```
adamconnelly, line 126: nit: I don't think `description` or `protectFromDeletion` are used anymore, so you could just inline this.

adamconnelly (approved):

spacelift/resource_module_test.go
cube2222: Good point, fixed.

PR #346, opened 2022-08-16
Add stack dependencies

peterdeme:
## Description of the change

Add stack dependencies.

## Type of change

## Related issues

Related to the Stack Dependencies feature.

### Development

- [ ] Lint rules pass locally

### Code review

- [ ] The target branch is `future` unless the change is going directly into production


spacelift/resource_stack_dependency.go
```diff
@@ -0,0 +1,152 @@
...
+			},
+			"depends_on_stack_id": {
+				Type:        schema.TypeString,
+				Description: "immutable ID (slug) of stack to depend on.",
+				Required:    true,
+				ForceNew:    true,
+			},
+			"triggers": {
+				Type:        schema.TypeBool,
+				Description: "describes whether we should trigger the dependent if it's not triggered by the push, but the current stack has changed. Defaults to `true`.",
```
cube2222, line 46: I'm not sold on this defaulting to true. Shouldn't this only be true in a very limited set of circumstances (after we have stack output references - only whenever a stack has an output that is passed out of band?)?

spacelift/resource_stack_dependency.go
```diff
@@ -0,0 +1,152 @@
...
+		StackDependency structs.StackDependency `graphql:"stackDependencyCreate(input: $input)"`
+	}
+
+	variables := map[string]interface{}{"input": stackDependencyCreateInput(d)}
+
+	if err := meta.(*internal.Client).Mutate(ctx, "StackDependencyCreate", &query, variables); err != nil {
+		return diag.Errorf("could not create stack dependency: %s", err)
+	}
+
+	d.SetId(fmt.Sprintf("%s|%s", query.StackDependency.ID, query.StackDependency.StackID))
```
cube2222, line 66: Why does this use the `|` character, instead of using `/` like the backend (and i.e. the policy attachment resource)?

peterdeme: because there is a slash in the dependency slug 😂

spacelift/resource_stack_dependency.go
michalg9: Hmm, I think I understand what you mean. My feeling is that while we don't have output references, default behaviour set to true makes more sense, whereas when having output references a default behaviour of false will make more sense.

After we have stack output references, the behaviour will be like that:
- if there was a change in outputs that are referenced in current stack -> behaviour similar to triggers:true
- if there were no changes -> behaviour similar to triggers:false -> unless we specified differently

Generally the idea with this field was to give users option to control how prefilling should work. Maybe we should actually not be setting defaults here, and require explicit definition? 🤔

peterdeme: We could make this mandatory without default value

cube2222: That would make sense to me.

spacelift/resource_stack_dependency.go
cube2222: whaaaat, how so?

cube2222: Slashes can be in slugs? Wouldn't that break the frontend URLs?

peterdeme: image

spacelift/resource_stack_dependency.go
```diff
@@ -0,0 +1,151 @@
...
+		StackDependency structs.StackDependency `graphql:"stackDependencyCreate(input: $input)"`
+	}
+
+	variables := map[string]interface{}{"input": stackDependencyCreateInput(d)}
+
+	if err := meta.(*internal.Client).Mutate(ctx, "StackDependencyCreate", &query, variables); err != nil {
+		return diag.Errorf("could not create stack dependency: %s", err)
+	}
+
+	d.SetId(path.Join(query.StackDependency.ID, query.StackDependency.StackID))
```
cube2222, line 65: I think in that case the more intuitive ordering would be `stackID/dependencyID`, like with policies, `policyID/attachmentID`

peterdeme: right, true

peterdeme: ✅

docs/resources/stack_dependency.md
```diff
@@ -0,0 +1,27 @@
...
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "spacelift_stack_dependency Resource - terraform-provider-spacelift"
+subcategory: ""
+description: |-
+  spacelift_stack_dependency represents a Spacelift stack dependency - a dependency between two stacks. When one stack depends on another, the tracked runs of the stack will not start until the dependent stack is successfully finished.
+---
+
+# spacelift_stack_dependency (Resource)
+
+`spacelift_stack_dependency` represents a Spacelift **stack dependency** - a dependency between two stacks. When one stack depends on another, the tracked runs of the stack will not start until the dependent stack is successfully finished.
```
cube2222, line 11: Additionally, changes to the dependency will trigger the dependent.

cube2222 (approved):

docs/resources/stack_dependency.md
peterdeme: good call, added

PR #430, opened 2023-03-01
Add secret headers for named webhooks

tomasmik:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]()

### Development

- [x] Lint rules pass locally

### Code review


spacelift/resource_named_webhook_secret_header.go
```diff
@@ -0,0 +1,152 @@
...
+	if err := meta.(*internal.Client).Mutate(ctx, "NamedWebhookSecretHeaderSet", &mutation, variables); err != nil {
+		ret = diag.Errorf("could not set secret header: %v", internal.FromSpaceliftError(err))
+	}
+
+	d.SetId(fmt.Sprintf("%s/%s", d.Get("webhook_id"), d.Get("key")))
+
+	return append(ret, resourceNamedWebhookSecretHeaderRead(ctx, d, meta)...)
+}
+
+func resourceNamedWebhookSecretHeaderRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
```
marcinwyszynski, line 87: If the changes happen behind the scenes (eg. manually changed in the GU), these will not be detected by Terraform. Within the current API I don't see a way of addressing it, and it's probably not a big deal in the 1st place, but in the future it would be great to see API design that is more Terraform-friendly.

spacelift/resource_named_webhook_secret_header.go
```diff
@@ -0,0 +1,152 @@
...
+	return nil
+}
+
+func resourceNamedWebhookSecretHeaderDelete(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	idParts := strings.SplitN(d.Id(), "/", 2)
+	if len(idParts) != 2 {
+		return diag.Errorf("unexpected resource ID: %s", d.Id())
+	}
+
+	var mutation struct {
```
marcinwyszynski, line 134: This will fail if the named webhook is gone, and this is not unusual because if you're deleting both things at the same time, the webhook will go 1st.

There's a few things we can do here:
- in a delete, perform a read 1st, and if the webhook is gone, call it a success - not ideal because it's not atomic/transactional;
- rescue an API error that will be reported when the webhook is gone, call it a success - not great, not terrible;
- change the mutation API to return a `nil` response when the webhook is gone, and change this code to take it into account;

marcinwyszynski (dismissed):

spacelift/resource_named_webhook_secret_header.go
tomasmik: Good point, will take that in to account in the future. It ended up being a strange API design all together tbh. In the end I'm not super happy about it either. :/

marcinwyszynski: It's not a high traffic resource so it's probably fine.

spacelift/resource_named_webhook_secret_header.go
tomasmik: > change the mutation API to return a nil response when the webhook is gone, and change this code to take it into account;

I don't think there is anything to take in to account here? Basically if it doesn't return an error, the code just continues and sets the ID to be empty without parsing anything extra.

spacelift/resource_named_webhook_secret_header.go
```diff
@@ -0,0 +1,152 @@
...
+				Required:         true,
+				ValidateDiagFunc: validations.DisallowEmptyString,
+				ForceNew:         true,
+			},
+			"value": {
+				Type:             schema.TypeString,
+				Description:      "value for the header",
+				DiffSuppressFunc: suppressValueChange,
+				Required:         true,
+				Sensitive:        true,
```
michieldewilde, line 50: value can be an empty string?

michieldewilde (dismissed):
one nit

spacelift/resource_named_webhook_secret_header.go
tomasmik: @michieldewilde, an empty string is a valid header value according to RFC 7230

While I would guess that almost no one will ever do that, it's better to allow this, rather than add artificial restrictions.

michieldewilde (approved):

PR #444, opened 2023-04-12
Add stack_activator resource to terraform provider

dobarx:
## Description of the change

Created a new resource for enabling/disabling Spacelift stacks. Example:

```hcl
resource "spacelift_stack" "app" {
  branch     = "master"
  name       = "Application stack"
  repository = "app"
}

resource "spacelift_stack_activator" "test" {
  enabled  = true
  stack_id = spacelift_stack.app.id
}

```

## Type of change

### Development

- [x] Lint rules pass locally

### Code review


Clickup-user:
Task linked: [CU-334hhwd Add `stack_activator` resource to terraform provider](https://app.clickup.com/t/334hhwd)

spacelift/resource_stack_activator.go
```diff
@@ -0,0 +1,129 @@
...
+	return nil
+}
+
+func resourceStackActivatorUpdate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	stack, err := queryStack(ctx, d, meta)
+	if err != nil {
+		return diag.Errorf("failed to retrieve stack: %v", internal.FromSpaceliftError(err))
+	}
+	if stack == nil {
+		d.SetId("")
```
adamconnelly, line 71: Should you not return an error if the stack is `nil`? I guess this would mean that either the stack doesn't exist or you don't have access to the stack. Either way I don't think enabling/disabling would work in that case.

spacelift/resource_stack_activator.go
```diff
@@ -0,0 +1,129 @@
...
+		d.SetId(fmt.Sprintf("activator-%d", time.Now().Unix()))
+	}
+	enabled, ok := d.Get("enabled").(bool)
+	if !ok {
+		return diag.Errorf("invalid enabled attribute: %v", d.Get("enabled"))
+	}
+	if !stack.IsDisabled == enabled {
+		return nil
+	}
+	if stack.IsDisabled {
```
adamconnelly, line 82: Personally I think it would make more sense if you used `enabled` here - that way we're basing the action on the Terraform property value, not on the current state of the stack. I know the outcome is identical either way, it just feels slightly wrong to me as-is.

dobarx:
@adamconnelly 
> So what if we flipped it slightly:
> 
> 1. Try to invoke the enable/disable mutation. I'm assuming it'll fail if the stack doesn't exist / you don't have access to it (but you'd need to double check that).
> 2. Call `resourceStackActivatorRead` to query for the stack, and set the state based on the `IsDisabled` field.
> 
> Just a thought - feel free to ignore me if that doesn't make sense for some reason.

But how do we know which mutation to invoke? We have two separate mutation queries `stackDisable(id)` & `stackEnable(id)`.
I do not know which to call unless I query stack first and compare.

adamconnelly:
> @adamconnelly
> 
> > So what if we flipped it slightly:
> > 
> > 1. Try to invoke the enable/disable mutation. I'm assuming it'll fail if the stack doesn't exist / you don't have access to it (but you'd need to double check that).
> > 2. Call `resourceStackActivatorRead` to query for the stack, and set the state based on the `IsDisabled` field.
> > 
> > Just a thought - feel free to ignore me if that doesn't make sense for some reason.
> 
> But how do we know which mutation to invoke? We have two separate mutations `stackDisable(id)` & `stackEnable(id)`. I do not know which to call unless I query stack first and compare. It would only be possible if we had something like `setStackEnabled(id, value)`.

I guess I figured you'd know which one to call based on the value of the `enabled` Terraform property. But maybe that doesn't handle the situation where the stack is enabled but the Terraform state thinks it's disabled or something like that?

adamconnelly:
@dobarx ok - no worries. Makes sense.

spacelift/resource_stack_activator.go
```diff
@@ -0,0 +1,128 @@
...
+	return resourceStackActivatorUpdate(ctx, d, meta)
+}
+
+func resourceStackActivatorRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	stack, err := queryStack(ctx, d, meta)
+	if err != nil {
+		return diag.Errorf("failed to retrieve stack: %v", internal.FromSpaceliftError(err))
+	}
+	if stack == nil {
+		d.SetId("")
```
adamconnelly, line 56: Maybe we should return an error here if the stack can't be found?

adamconnelly (commented):
Just noticed one other small thing. Other than that, looks good!

peterdeme (approved):

PR #459, opened 2023-08-04
Added spacelift_stack_dependency_reference resource

truszkowski:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]()

### Development

- [ ] Lint rules pass locally

### Code review


docs/resources/stack_dependency_reference.md
```diff
@@ -0,0 +1,51 @@
...
+
+resource "spacelift_stack_dependency" "test" {
+  stack_id            = spacelift_stack.app.id
+  depends_on_stack_id = spacelift_stack.infra.id
+}
+
+resource "spacelift_stack_dependency_reference" "test" {
+  stack_dependency_id = spacelift_stack_dependency.test.id
+  output_name         = "DB_CONNECTION_STRING"
+  input_name          = "APP_DB_URL"
```
adamconnelly, line 36: Do we have a way to mark the dependencies as secrets? I'm just thinking for something like this you wouldn't want it being connected as plaintext, and then accidentally leaked.

spacelift/resource_stack_dependency_reference.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if err := meta.(*internal.Client).Mutate(ctx, "StackDependenciesDeleteReference", &query, variables); err != nil {
+		return diag.Errorf("could not delete stack dependency reference: %s", err)
+	}
+
+	d.SetId("")
+	return nil
+}
+
+func getStackDependencyIDParts(d *schema.ResourceData) ([]string, diag.Diagnostics) {
```
adamconnelly, line 161: Really minor suggestion, but what about instead of returning a slice, return the components. That way it's really easy to understand the calling code because you'd be able to write something like this:

```go
stackID, stackDependencyID, diags := getStackDependencyIDParts(d)
```

spacelift/resource_stack_dependency_reference_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+
+	. "github.com/spacelift-io/terraform-provider-spacelift/spacelift/internal/testhelpers"
+)
+
+func TestStackDependencyReferenceResource(t *testing.T) {
+	const resourceName = "spacelift_stack_dependency_reference.test"
+
+	t.Run("creates and updates stack dependency reference", func(t *testing.T) {
```
adamconnelly, line 16: Could you add a test for deletion as well?

adamconnelly (commented):
Looks good, although looks like you need to rebase and also fix a linting error.

docs/resources/stack_dependency_reference.md
truszkowski: Which variable is sensitive is defined in an output configuration. Reference is only a link output -> input. 
```tf
output "my_secret_output" {
    value = ...
    sensitive = true
}
```

adamconnelly: Cool - makes sense.

spacelift/resource_stack_dependency_reference.go
```diff
@@ -0,0 +1,179 @@
...
+			InputName:  toString(d.Get("input_name")),
+			Type:       toString("ENVIRONMENT_VARIABLE"),
+		},
+	}
+
+	if err := meta.(*internal.Client).Mutate(ctx, "StackDependenciesAddReference", &query, variables); err != nil {
+		return diag.Errorf("could not create stack dependency reference: %s", err)
+	}
+
+	d.SetId(path.Join(idParts[0], idParts[1], query.StackDependencyReference.ID))
```
peterdeme, line 74: hm, I think we have reference ULIDs, right? That'd be a good ID.

truszkowski: Yes, `query.StackDependencyReference.ID` is ULID. But we also need `stack_dependency_id` and `stack_id`.
We querying: `stack($stack_id) { dependency($stack_dependency_id) { reference($id) { ... } } } }`. 
Which is similar to querying dependency (but one layer less). 

We could add a direct query (like `stack_dependency_reference($id)` without nested queries), but I saw here many places where we use this id format (`idPart[0]/idPart[1]/...`). 

Btw, we also verify a stack or dependency exists; a stack contains dependency, etc... Skipping stack_id and dependency_id, we can't do these checks.

peterdeme: ah nevermind, right

spacelift/resource_stack_dependency_reference_test.go
truszkowski: added

peterdeme (approved):

PR #473, opened 2023-10-04
Add support for user group resources

mwasilew2:
### User need

User Management is a feature that is in closed beta right now. Most of it is already implemented, but there are still some bits and pieces missing. One example of a missing bit is that the terraform provider is missing support for managing resources related to User Management. This PR is intended to improve on that by adding support for managing user group resources.

### Solution

Add support to the spacelift terraform provider for managing User Management related objects.

## Type of change

- [x] Documentation (non-breaking change that adds documentation)

### Development

- [x] Lint rules pass locally

### Code review


spacelift/internal/structs/user_group.go
```diff
@@ -0,0 +1,12 @@
+package structs
+
+type SpaceAccessRule struct {
+	Space            string `graphql:"space"`
+	SpaceAccessLevel string `graphql:"spaceAccessLevel"`
+}
+
+type UserGroup struct {
```
mwasilew2, line 8: I'm wondering if perhaps a different name here would be more meaningful, for example, `UserGroupBinding`

spacelift/internal/structs/user_group.go
```diff
@@ -0,0 +1,12 @@
+package structs
+
+type SpaceAccessRule struct {
```
mwasilew2, line 3: In a similar spirit, this could be named `Role`

mwasilew2:
The tests pass against my local instance. My current suspicion is that they fail in Github Actions because the API which the tests run against doesn't include the relevant graphql schema changes.

The provider uses `SPACELIFT_API_KEY_ENDPOINT` . I wanted to look it up in the repo config, but all env vars are stored as secrets so there's no way to check it that way.

mwasilew2:
@adamconnelly @truszkowski @marcinwyszynski 

I'm troubleshooting: https://github.com/spacelift-io/terraform-provider-spacelift/actions/runs/6407726212/job/17395133360?pr=473

I noticed you've been recently active in this repo, do you know which  instance is used for running tests? (preprod, prod, another one?)

adamconnelly:
> @adamconnelly @truszkowski @marcinwyszynski
> 
> I'm troubleshooting: https://github.com/spacelift-io/terraform-provider-spacelift/actions/runs/6407726212/job/17395133360?pr=473
> 
> I noticed you've been recently active in this repo, do you know which instance is used for running tests? (preprod, prod, another one?)

@mwasilew2 I think the issue will just be that the new user management feature flags haven't been enabled for the test account (which is one of our preprod accounts). I'll message you on Slack with more info.

mwasilew2:
on preprod I got this:

Screenshot 2023-10-04 at 18 04 16

spacelift/internal/structs/user_group.go
michieldewilde: The concept of a Role within user management already exists.
The role can be admin, owner, and user, so this name is not available.

spacelift/internal/structs/user_group.go
michieldewilde: In the UI, we are using the name IdP group mapping.

spacelift/resource_user_group.go
```diff
@@ -0,0 +1,176 @@
...
+	"ADMIN",
+}
+
+func resourceUserGroup() *schema.Resource {
+	return &schema.Resource{
+		Description: "" +
+			"`spacelift_user_group` represents a Spacelift **user group** - " +
+			"a collection of users as provided by your Identity Provider (IdP). " +
+			"If you assign permissions to a user group, all users in the group " +
+			"will have those permissions, unless the user's permissions are higher than " +
```
michieldewilde, line 28: > `permissions, unless`
It does not need a comma.

michieldewilde:
I think we are missing a source for user groups?
Or will that be in a separate PR?

spacelift/internal/structs/user_group.go
michieldewilde, line 8: There should be a field that indicates the IdP for this user group.

michieldewilde, line 8: The user group should have a status (inactive or active).

michieldewilde (changes requested):
A few fields could be included that would be useful. 
I'm not sure if there was an intentional decision not to add a datasource for the user groups. If there was could you please elaborate?

spacelift/internal/structs/user_group.go
marcinwyszynski: The IdP is not controlled by the user, and it would be a computed fueld.

I think it will only make more sense to add more of these fields if we make this a data source. In the 1st version of the resource I'd suggest focusing on things that users need to manage access.

marcinwyszynski: Same as above.

michieldewilde: That's fine to me.

michieldewilde: That's fine with me.

michieldewilde (approved):

spacelift/internal/structs/user_group.go
mwasilew2: in that case I'm not gonna introduce a third name here, I'll stick with what's used in the GraphQL schema

spacelift/internal/structs/user_group.go
mwasilew2: similar to the other one, I'll stick with what's in the api definition

spacelift/internal/structs/user_group.go
michieldewilde: The problem in that case is that the UI and terraform provider would have different names. They are meant to be similar, at least based on naming, because the person implementing the terraform provider will search for resources that match the naming in the UI.
@marcinwyszynski, what do you think from a usability/conformity perspective?

marcinwyszynski: @michieldewilde This is a struct representing the GraphQL `type ManagedUserGroup`. If we want consistency, this is the "canonical" name.

marcinwyszynski:
@michieldewilde I explicitly requested that the data sources are handled separately.

spacelift/internal/structs/user_group.go
michieldewilde: I'm referring to the name of the resource (commented in the wrong place).
There is a resource with the name `spacelift_user_group`; we are not using the name user group in any other place than the GraphQL API.
Searching in the documentation and our UI, I can't find any reference to a user group.
We are calling it IdP Group Mapping in those places.
This might confuse users.
There will be essentially two different names for the same concept in public-facing systems (three if you count the difference between `UserGroup` and `ManagedUserGroup`).

marcinwyszynski: OK, `spacelift_idp_group_mapping` then?

michieldewilde: Yes.
It would be best to do it this way and add a description to the `ManagedUserGroup` in GraphQL to clarify that this refers to the IdP group mapping in the UI and the terraform provider.

mwasilew2:
It's not pretty, but the user facing interface is consistent with the UI and docs.

docs/resources/idp_group_mapping.md
```diff
@@ -0,0 +1,47 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "spacelift_idp_group_mapping Resource - terraform-provider-spacelift"
+subcategory: ""
+description: |-
+  spacelift_idp_group_mapping represents a mapping (binding) between a user group (as provided by IdP) and a Spacelift User Management Policy. If you assign permissions (a Policy) to a user group, all users in the group will have those permissions unless the user's permissions are higher than the group's permissions.
```
michieldewilde, line 6: I wouldn't use Spacelift User Management Policy.
```suggestion
  spacelift_idp_group_mapping represents a mapping (binding) between a user group (as provided by IdP) and a set of access management rules. All users in the group will have those permissions unless the user's permissions are higher than the group's permissions.
```

spacelift/resource_idp_group_mapping.go
```diff
@@ -0,0 +1,174 @@
...
+var validAccessLevels = []string{
+	"READ",
+	"WRITE",
+	"ADMIN",
+}
+
+func resourceIdpGroupMapping() *schema.Resource {
+	return &schema.Resource{
+		Description: "" +
+			"`spacelift_idp_group_mapping` represents a mapping (binding) between a user group (as provided by IdP) " +
```
michieldewilde, line 25: > spacelift_idp_group_mapping represents a mapping (binding) between a user group (as provided by IdP) and a set of access management rules. All users in the group will have those permissions unless the user's permissions are higher than the group's permissions.

Spacelift User Management Policy sounds too much like a login policy.

michieldewilde (approved):

PR #497, opened 2024-01-10
feat(security-email): add support for security email

mwasilew2:
## Description of the change

related to: https://app.clickup.com/t/8693fq8dx

Adds to the terraform provider support for editing security email.

Note: the graphql api doesn't currently support deleting security email and terraform test sdk runs delete after each test

## Type of change

## Related issues

> Fix [#1]()

### Development

- [ ] Lint rules pass locally

### Code review


Clickup-user:
Task linked: [CU-8693fq8dx add support for `security_email` to terraform provider](https://app.clickup.com/t/8693fq8dx)

mwasilew2:
similar to: https://github.com/spacelift-io/terraform-provider-spacelift/pull/495

spacelift/resource_security_email.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	if err := i.(*internal.Client).Mutate(ctx, "AccountUpdateSecurityEmail", &mutation, variables); err != nil {
+		return diag.Errorf("could not create security email: %v", err)
+	}
+
+	return resourceSecurityEmailRead(ctx, data, i)
+}
+
+func resourceSecurityEmailDelete(ctx context.Context, data *schema.ResourceData, i interface{}) diag.Diagnostics {
+	return diag.Errorf("deleting security email is not supported")
```
adamconnelly, line 86: This doesn't seem a great UX. Wouldn't it mean that if you had added this resource, you'd need to remove it from the Terraform state in order to be able to delete the resource?

I wonder if we should do something like this:

1. Update the API to allow the security email to be set to an empty string / null; or
2. Turn this into a no-op (maybe with a warning).

spacelift/resource_security_email.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
+	"github.com/spacelift-io/terraform-provider-spacelift/spacelift/internal"
+)
+
+func resourceSecurityEmail() *schema.Resource {
+	return &schema.Resource{
+		Description: "" +
+			"`spacelift_security_email` represents a security email address that " +
```
adamconnelly, line 16:
```suggestion
			"`spacelift_security_email` represents an email address that " +
```

Just a nit - I don't think we need `security` in here twice.

adamconnelly (commented):
I'm not convinced about the delete behaviour, but other than that it makes sense.

Can you add a test for the resource please?

spacelift/resource_security_email.go
mwasilew2: > you'd need to remove it from the Terraform state in order to be able to delete the resource?

yes

I'll turn it into a warning and remove the resource from the state

adamconnelly (approved):

PR #499, opened 2024-01-15
added resources for saved filter handling

mplushnikov:
## Description of the change

Added resource and datasource for handling of spacelift saved filter, E.g. to be able to add predefined filters for different teams

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation (non-breaking change that adds documentation)

### Development

- [x] Lint rules pass locally
- [x] The code changed/added as part of this pull request has been covered with tests
- [x] All tests related to the changed code pass in development
- [x] Examples for new resources and data sources have been added
- [x] Default values have been documented in the description (e.g., "Dummy: (Boolean) Blah blah. Defaults to `false`.)
- [x] If the action fails that checks the documentation: Run `go generate` to make sure the docs are up to date

### Code review

- [ ] This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [ ] Pull Request is no longer marked as "draft"
- [ ] Reviewers have been assigned
- [ ] Changes have been reviewed by at least one other engineer

PR #521, opened 2024-03-04
added resources for saved filter handling

truszkowski:
## Description of the change

Copied from: [PR](https://github.com/spacelift-io/terraform-provider-spacelift/pull/499). GitHub Action secrets don't work with forks.

Added resource and data source for handling of spacelift saved filter, E.g. to be able to add predefined filters for different teams.

## Type of change

## Related issues

> Fix [#1]()

### Development

- [ ] Lint rules pass locally

### Code review

- [x] Changes have been reviewed by at least one other engineer


spacelift/data_saved_filter.go
```diff
@@ -0,0 +1,85 @@
...
+		} `graphql:"savedFilter(id: $id)"`
+	}
+
+	variables := map[string]interface{}{"id": d.Get("filter_id")}
+
+	if err := meta.(*internal.Client).Query(ctx, "savedFilter", &query, variables); err != nil {
+		return diag.Errorf("could not query for filter: %v", err)
+	}
+
+	d.SetId(query.Filter.ID)
```
adamconnelly, line 77: Does this handle the situation where the filter doesn't exist? I think the API will return `nil` in that case rather than an error, and presumably we'll just end up with an empty `Filter` object.

spacelift/data_saved_filters.go
```diff
@@ -0,0 +1,110 @@
...
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataFiltersRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
+	var query struct {
+		Filters []structs.SavedFilter `graphql:"savedFilters()"`
```
adamconnelly, line 76: It looks like the API supports querying by type (`savedFilters(type: String)`). Maybe we should use that and only do the name filtering client-side?

spacelift/data_saved_filters_test.go
```diff
@@ -0,0 +1,40 @@
...
+	. "github.com/spacelift-io/terraform-provider-spacelift/spacelift/internal/testhelpers"
+)
+
+func TestFiltersData(t *testing.T) {
+	t.Run("load all saved filters", func(t *testing.T) {
+		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
+
+		datasourceName := "data.spacelift_saved_filters.all"
+
+		testSteps(t, []resource.TestStep{{
```
adamconnelly, line 19: I think we could do with a few more tests here - testing for each filter, and also for no results maybe?

adamconnelly (commented):
Looks really good. Left some comments, but nothing major.

spacelift/data_saved_filter.go
truszkowski: You're right; I'll fix that and add a test for that case.

truszkowski: Fixed.

spacelift/data_saved_filters_test.go
truszkowski: Added more tests.

adamconnelly (approved):
Nice one!

eliecharra (approved):

Issue #427, opened 2023-02-08
Add tail option to spacelift_run resource


jwenz723:
I have a terraform module which creates stacks in spacelift using the `spacelift_stack` terraform resource. I am trying to create an automated test for my module using the spacelift module registry. The test runs my module, which causes the child stacks to be created, but it does not automatically run the child stacks. In my test I would like to run the child stacks so that I can validate they are configured properly.

I am able to use the `spacelift_run` resource to trigger the child stacks to run, but there is not a way for me to know when the run has completed. I need to know when the run has completed so that my automated test can destroy the child stacks.

Yantrio:
This was added in #535 and released in v1.11.0 https://github.com/spacelift-io/terraform-provider-spacelift/releases/tag/v1.11.0

PR #522, opened 2024-03-06
feat: wait for run in resource `spacelift_run`

tmeckel:
## Description of the change

This PR contains an extension of the resource `spacelift_run`, which makes it possible to wait for the completion of a run.

## Type of change

## Related issues

Closes: #427

### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/docs/resources/run.md
+++ b/docs/resources/run.md
@@ -40,7 +40,27 @@ resource "spacelift_run" "this" {
 - `commit_sha` (String) The commit SHA for which to trigger a run.
 - `keepers` (Map of String) Arbitrary map of values that, when changed, will trigger recreation of the resource.
 - `proposed` (Boolean) Whether the run is a proposed run. Defaults to `false`.
+- `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
+- `wait` (Block List, Max: 1) Wait for the run to finish (see [below for nested schema](#nestedblock--wait))
 
 ### Read-Only
 
 - `id` (String) The ID of the triggered run.
+
+<a id="nestedblock--timeouts"></a>
+### Nested Schema for `timeouts`
+
+Optional:
+
+- `create` (String)
+
+
+<a id="nestedblock--wait"></a>
+### Nested Schema for `wait`
+
+Optional:
+
+- `continue_on_discarded` (Boolean) Continue if run has been discarded. Default: `false`
+- `continue_on_failed` (Boolean) Continue if run failed. Default: `false`
+- `continue_on_timeout` (Boolean) Continue if run timed out, i.e. did not reach end state `finished` in time. Default: `false`
+- `enabled` (Boolean) Whether waiting for a job is enabled or not. Default: `false`
--- a/docs/resources/run.md
+++ b/docs/resources/run.md
@@ -62,4 +62,4 @@ Optional:
 
 - `continue_on_state` (Set of String) Continue on the specified states of a finished run. If not specified, the default is `[ 'finished' ]`
 - `continue_on_timeout` (Boolean) Continue if run timed out, i.e. did not reach any defined end state in time. Default: `false`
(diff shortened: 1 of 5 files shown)


docs/resources/run.md
```diff
@@ -40,7 +40,27 @@ resource "spacelift_run" "this" {
...
+
+- `create` (String)
+
+
+<a id="nestedblock--wait"></a>
+### Nested Schema for `wait`
+
+Optional:
+
+- `continue_on_discarded` (Boolean) Continue if run has been discarded. Default: `false`
```
marcinwyszynski, line 63: Would it make sense to create a single `[]string` field like `continue_on` where users can add any final state they want?

tmeckel: @marcinwyszynski as requested, now using `continue_on_state` which accepts a list of possible (acceptable) end states. The `finished` state is the default state to wait for. I doubt Spacelift will change that name any soon.

spacelift/resource_run.go
```diff
@@ -80,11 +145,149 @@ func resourceRunCreate(ctx context.Context, d *schema.ResourceData, meta interfa
...
 
+	if waitRaw, ok := d.GetOk("wait"); ok {
+		wait := expandWaitConfiguration(waitRaw.([]interface{}))
+
+		if wait.enabled {
+			stateConf := &retry.StateChangeConf{
+				ContinuousTargetOccurence: 1,
+				Delay:                     10 * time.Second, // TODO: Delay must be a multiple of Timeout
+				MinTimeout:                10 * time.Second,
+				Pending: []string{
```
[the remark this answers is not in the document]
tmeckel: @marcinwyszynski using `finished: Boolean!` in `checkStackStatusFunc` for `stateConf.WaitForStateContext`

spacelift/resource_run.go
```diff
@@ -55,16 +67,78 @@ func resourceRun() *schema.Resource {
...
 			},
+			"wait": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Description: "Wait for the run to finish",
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:        schema.TypeBool,
```
marcinwyszynski, line 78: Naive question perhaps but do we need it? Or would the presence of the `wait` block clearly indicate that this is `true`?

tmeckel: @marcinwyszynski this gives practitioners the ability to switch waiting on and of by using a variable instead of a dynamic block. Much easier and cleaner to my taste.

marcinwyszynski: Yeah that makes sense. But most of the time the intention would be to have it enabled, no? Just wondering then if it could be better UX if the logic was flipped - so the field would be called `disabled`, and `false` by default? I'm slightly worried that if someone forgets to put the `enabled: true` line, the behavior of the resource would fail their expectations.

tmeckel: @marcinwyszynski Fair objection. Changed it according you proposal

spacelift/resource_run.go
```diff
@@ -80,11 +154,121 @@ func resourceRunCreate(ctx context.Context, d *schema.ResourceData, meta interfa
...
+		}
+		state = "running"
+		if finished {
+			state = "finished"
+		}
+		return
+	}
+}
+
+func getStackRunStateByID(ctx context.Context, client *internal.Client, stackID string, runID string) (string, bool, error) {
```
marcinwyszynski, line 246: Actually, I don't think this is going to work inside Spacelift.

The tokens that Spacelift runs get are more limited in scope than regular tokens and they will not allow you to make this query. I think this will require some investigation on our end - perhaps a very limited endpoint to get run status by ID?

marcinwyszynski: I created an internal ticket to support this on the backend.

tmeckel:
@marcinwyszynski @truszkowski closing my PR in favor of #535

PR #536, opened 2024-04-11
feat: add support for choosing the Terragrunt tool

adamconnelly:
## Description of the change

- Added a new `tool` field to allow users to choose between OpenTofu, Terraform, or manually provisioning the tool themselves.
- Fixed an existing issue in the stack data source and resource where we weren't specifying the vendor config for Terragrunt stacks properly.
- Related to the previous point, we were treating Terragrunt stacks like Terraform stacks, and setting all the `terraform_` settings. But the API actually returns `null` from the `vendorConfig.terraform` property for Terragrunt stacks, so these would have all contained the default values for their types (i.e. `null`, `false`, etc).
- Fixed a bug where we were passing through empty strings instead of `null` when the `terraform_version` or `terragrunt_version` properties weren't specified. The API actually handles it fine, but I figured it made sense to fix.

## Type of change

### Development

- [x] Lint rules pass locally

### Code review


Clickup-user:
Task linked: [CU-86941f845 [Terragrunt-OpenTofu] Add support for managing stacks via Terraform provider](https://app.clickup.com/t/86941f845)

examples/resources/spacelift_stack/resource.tf
```diff
@@ -164,3 +164,21 @@ resource "spacelift_stack" "ansible-stack" {
   repository   = "ansible-playbooks"
   runner_image = "public.ecr.aws/spacelift/runner-ansible:latest"
 }
+
+# Terragrunt stack using github.com as VCS
+resource "spacelift_stack" "terragrunt-stack" {
+  terragrunt {
+    terraform_version      = "1.6.2"
```
peterdeme, line 171: are we sure we want to call it terraform version?

adamconnelly: It's not a new field, so I can't really change it without breaking backwards compatibility. I could introduce a new field and deprecate it, but I'd rather not add that complication.

peterdeme: yeah but wouldn't it make more sense to add an open tofu version?

- if both populated thrown an error
- if opentofu version populated, set the tool to tofu
- tf version is populated, set the tool to tf

adamconnelly: I don't think that would work for a couple of reasons:

1. We support defaulting to the latest version if you don't specify it. So we'd still need a separate `tool` field to indicate what you wanted to use in that situation.
2. We support manual provisioning, where no OpenTofu / Terraform version is specified.

adamconnelly: We could add an `opentofu_version` field or `tool_version`, and just allow either to be used. Not sure I love having two fields for the same thing though :disappointed:

peterdeme: keep it as is

eliecharra (approved):
Nice one 👍🏻

spacelift/data_stack.go
```diff
@@ -444,6 +444,40 @@ func dataStack() *schema.Resource {
...
+				Type:        schema.TypeList,
+				Description: "Terragrunt-specific configuration. Presence means this Stack is a Terragrunt Stack.",
+				Computed:    true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"terraform_version": {
+							Type:        schema.TypeString,
+							Description: "The Terraform version.",
+							Computed:    true,
+						},
```
truszkowski, line 457: Should we set `Optional: true` here? Or Default?

adamconnelly: I don't think so, because it's the data-source, not the resource. So it'll always be populated.

truszkowski: Ah, it's the data source, not a resource. Nevermind.

truszkowski (approved):

PR #556, opened 2024-05-17
Add custom headers for audit trail

ilya-hontarau:
## Description of the change

> Description here

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review


spacelift/resource_audit_trail_webhook.go
```diff
@@ -47,7 +47,12 @@ func resourceAuditTrailWebhook() *schema.Resource {
 				Type:        schema.TypeString,
 				Required:    true,
 				Sensitive:   true,
-				Description: "`secret` is a secret that Spacelift will send with the request",
+				Description: "`secret` is a secret that Spacelift will send with the request.",
+			},
+			"custom_headers": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				Description: "`custom_headers` is a Map of key values strings, that will be passed as headers with audit trail call.",
```
adamconnelly, line 55: I wonder if this should be "with the audit trail call", or "with audit trail calls"? Maybe we should use the word "request" instead of "call" as well?

adamconnelly (commented):
Looks good. Just to check one other thing - say I want to send something like an API key in a header, can I use the `sensitive()` function to mark the header value as sensitive, and it'll be correctly masked in the Terraform logs and in our resource changes view?

Probably a stupid question, but figured I'd ask.

ilya-hontarau:
image
image

adamconnelly (approved):

ilya-hontarau:
I think we are going to change API a bit, so possible this client will be changed

spacelift/internal/structs/audit_trail_webhook.go
```diff
@@ -1,6 +1,12 @@
 package structs
 
 type AuditTrailWebhook struct {
+	AuditTrailWebhookRead
+	// CustomHeaders is used only for create/update, API doesn't return it back.
```
eliecharra, line 5:
```suggestion
	// CustomHeaders is used only for create/update, API doesn't return them back.
```

eliecharra (dismissed):

adamconnelly (approved):
It looks good - I'd maybe be tempted to hold off merging until the frontend changes are made to support this. Otherwise it could be a little confusing because you'd be able to configure this via TF, but wouldn't be able to see that the headers had been configured anywhere in the UI.

ilya-hontarau:
lets close it for now

PR #564, opened 2024-08-12
feat(env-vars): Add `description` field to Terraform spacelift_environment_variable resource

0michalsokolowski0:
## Description of the change

> Adding a new optional field to spacelift_environment_variable Terraform module called **description**, [more details](https://app.clickup.com/t/8695950mj).

image
image

## Type of change

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/spacelift/resource_environment_variable.go
+++ b/spacelift/resource_environment_variable.go
@@ -82,6 +82,12 @@ func resourceEnvironmentVariable() *schema.Resource {
 				Default:     true,
 				ForceNew:    true,
 			},
+			"description": {
+				Type:        schema.TypeString,
+				Description: "Description of the environment variable",
+				Optional:    true,
+				ForceNew:    true,
+			},
 		},
 	}
 }
@@ -89,10 +95,11 @@ func resourceEnvironmentVariable() *schema.Resource {
 func resourceEnvironmentVariableCreate(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
 	variables := map[string]interface{}{
 		"config": structs.ConfigInput{
-			ID:        toID(d.Get("name")),
-			Type:      structs.ConfigType("ENVIRONMENT_VARIABLE"),
-			Value:     toString(d.Get("value")),
-			WriteOnly: graphql.Boolean(d.Get("write_only").(bool)),
+			ID:          toID(d.Get("name")),
+			Type:        structs.ConfigType("ENVIRONMENT_VARIABLE"),
+			Value:       toString(d.Get("value")),
+			WriteOnly:   graphql.Boolean(d.Get("write_only").(bool)),
+			Description: toOptionalString(d.Get("description")),
 		},
 	}
 
@@ -202,6 +209,12 @@ func resourceEnvironmentVariableRead(ctx context.Context, d *schema.ResourceData
 		d.Set("value", *value)
 	} else {
 		d.Set("value", element.Checksum)
+	}
+
+	if description := element.Description; description != nil {
+		d.Set("description", *description)
+	} else {
+		d.Set("description", nil)
 	}
 
 	return nil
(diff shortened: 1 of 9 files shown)


spacelift/resource_environment_variable.go
```diff
@@ -82,17 +82,25 @@ func resourceEnvironmentVariable() *schema.Resource {
 				Default:     true,
 				ForceNew:    true,
 			},
+			"description": {
+				Type:             schema.TypeString,
+				Description:      "Description of the environment variable",
+				Optional:         true,
+				ValidateDiagFunc: validations.DisallowEmptyString,
+				ForceNew:         true,
```
0michalsokolowski0: Added to fix:
```
Error: InternalValidate
        
          with provider["registry.terraform.io/hashicorp/spacelift"],
          on terraform_plugin_test.tf line 1, in provider "spacelift":
           1: provider "spacelift" {}
        
        Internal validation of the provider failed! This is always a bug
        with the provider itself, and not a user issue. Please report
        this bug:
        
        resource spacelift_environment_variable: No Update defined, must set ForceNew
        on: []string{"description"}
```

test/main.tf
```diff
@@ -0,0 +1,46 @@
+terraform {
```
tomasmik, line 1: uuuuum, I think you included this by accident?

0michalsokolowski0: Yeah

spacelift/resource_environment_variable.go
```diff
@@ -82,17 +82,24 @@ func resourceEnvironmentVariable() *schema.Resource {
 				Default:     true,
 				ForceNew:    true,
 			},
+			"description": {
+				Type:        schema.TypeString,
+				Description: "Description of the environment variable",
+				Optional:    true,
+				ForceNew:    true,
```
mbialon, line 89: Recreating an environment variable will clear out the value that might have been set through the UI. It may not be evident to users why that's required. It may also require the user to update a secret value, which can be a headache.

0michalsokolowski0: Should I implement the **UpdateContext** for this field? It seems all other fields in this resource have this property true as well.

0michalsokolowski0: > Recreating an environment variable will clear out the value that might have been set through the UI.

If an environment variable was created with UI, Terraform won't track it. Users could only import it as data, which is read-only unless I am missing something.

mbialon: > > Recreating an environment variable will clear out the value that might have been set through the UI.
> 
> If an environment variable was created with UI, Terraform won't track it. Users could only import it as data, which is read-only unless I am missing something.

Users sometimes create the env variable through IaC but use the UI to set the value of that variable.

mbialon: > Should I implement the **UpdateContext** for this field?

I believe so. WDYT @tomasmik @Yantrio @silviutanasa?

> It seems all other fields in this resource have this property true as well.

It makes sense to recreate the variable when moving a variable between entities, setting the value, or changing a plain value into a secret. I believe it would be better if we didn't require a recreate when working with a description.

0michalsokolowski0: In the UI (modification of env var for contexts), the only way to save changes is to change value (neither making as secret or description change can be saved on their own).
Screenshot 2024-08-13 at 10 16 16

If the value is modified, the resource will get [recreated](https://github.com/spacelift-io/terraform-provider-spacelift/blob/6ef7b8791f3ad38628a0a6209324fdb78c4fa66b/spacelift/resource_environment_variable.go#L76).

mbialon: Recreating the variable makes it consistent with the UI.

mbialon (approved):
Good job!

tomasmik (approved):
😎

Yantrio (approved):

PR #568, opened 2024-09-18
Update provider.go to handle slash at api_key_endpoint

akmal-spacelift:
## Description of the change

> Handles the slash at api_key_endpoint

## Type of change

## Related issues

> internal ref: https://spacelift-io.slack.com/archives/C06L3C9L4P6/p1716303659139979?thread_ts=1716303553.805579&cid=C06L3C9L4P6 / https://spacelift-io.slack.com/archives/C03THR9LFH7/p1715837109837649?thread_ts=1715836160.097879&cid=C03THR9LFH7 / https://spacelift-io.slack.com/archives/C01724WEYJF/p1726588817203019

### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/spacelift/provider.go
+++ b/spacelift/provider.go
@@ -225,7 +225,15 @@ func buildClientFromToken(token string) (*internal.Client, error) {
 
 func buildClientFromAPIKeyData(d *schema.ResourceData) (*internal.Client, error) {
 	// Since validation runs first, we can safely assume that the data is there.
-	endpoint := fmt.Sprintf("%s/graphql", d.Get("api_key_endpoint").(string))
+	
+	// Handle / at api_key_endpoint
+	endpoint := d.Get("api_key_endpoint").(string)
+	if endpoint[len(api_key_endpoint)-1:] == "/" {
+		endpoint := fmt.Sprintf("%sgraphql", endpoint)
+	} else {
+		endpoint := fmt.Sprintf("%s/graphql", endpoint)
+	}
+	
 	apiKeyID := d.Get("api_key_id").(string)
 	apiKeySecret := d.Get("api_key_secret").(string)
 
--- a/spacelift/provider.go
+++ b/spacelift/provider.go
@@ -228,7 +228,7 @@ func buildClientFromAPIKeyData(d *schema.ResourceData) (*internal.Client, error)
 	
 	// Handle / at api_key_endpoint
 	endpoint := strings.TrimSuffix(d.Get("api_key_endpoint").(string), "/")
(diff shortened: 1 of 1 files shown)


akmal-spacelift:
updated https://github.com/spacelift-io/terraform-provider-spacelift/pull/568/files#diff-6e7375fac891c18ad89da0bd68e1e4993c044e8eaad6aa4926f1e3a14b4b0f1b

spacelift/provider.go
```diff
@@ -225,7 +225,15 @@ func buildClientFromToken(token string) (*internal.Client, error) {
 
 func buildClientFromAPIKeyData(d *schema.ResourceData) (*internal.Client, error) {
 	// Since validation runs first, we can safely assume that the data is there.
-	endpoint := fmt.Sprintf("%s/graphql", d.Get("api_key_endpoint").(string))
+	
+	// Handle / at api_key_endpoint
+	endpoint := d.Get("api_key_endpoint").(string)
+	if endpoint[len(api_key_endpoint)-1:] == "/" {
```
adamconnelly, line 231: This would need to be `len(endpoint)` - there's no `api_key_endpoint` variable.

But I think I'd go with `strings.TrimSuffix` here and then you don't need to worry about indexing. So something like:

```go
endpoint := strings.TrimSuffix(d.Get("api_key_endpoint").(string), "/")
endpiont = fmt.Sprintf("%s/graphql", endpoint)
```

Also, can you test this locally please and confirm that it works with both cases? We have info in the readme about building a local copy of the provider and testing it: https://github.com/spacelift-io/terraform-provider-spacelift?tab=readme-ov-file#using-a-local-build-of-the-provider.

adamconnelly (commented):
Thanks for the contribution @akmal-spacelift! It looks like there's a slight issue with the code causing it to not compile. I've left some suggestions.

eliecharra (commented):
What Adam said 😛

spacelift/provider.go
akmal-spacelift: Definitely!  Thanks Adam

akmal-spacelift: I've tested it locally both with and without the / at the end and it worked.  I've also added a // at the end and it failed as expected.

adamconnelly:
Thanks @akmal-spacelift for fixing this! I've opened a new PR with your changes (the issue is that some of our PR checks don't run if you don't have write access to the repo) here: https://github.com/spacelift-io/terraform-provider-spacelift/pull/568.

I'm going to close this PR in favour of that one.

Issue #549, opened 2024-05-07
Feature Request: Enhance spacelift_user Resource for Granular Policy Management


HighwayofLife:
### Description:
Currently, the `spacelift_user` resource in the Spacelift Terraform provider utilizes a repeating policy block to manage user role assignments across spaces. This design presents a major limitation: it requires importing and managing ALL of a user's policy assignments within a single Terraform configuration. Consequently, adding a user to a new space without modifying existing assignments becomes cumbersome and error-prone.

This feature request proposes an enhancement to the `spacelift_user` resource to enable granular policy management. This would allow adding or removing individual policy assignments without affecting existing ones in different spaces.

### Problem:
The current policy block structure forces users to manage all role assignments for a user within a single Terraform configuration. This approach presents several challenges:

- Increased Complexity: Managing all policy assignments centrally can lead to complex Terraform configurations, especially for users with roles in multiple spaces.
- Risk of Accidental Modifications: Modifying a user's policy in one space might unintentionally impact their roles in other spaces.
- Limited Flexibility: Adding a user to a new space without affecting existing assignments becomes difficult and requires workarounds.

### Proposed Solution:
Introduce a new resource, such as `spacelift_user_policy`, dedicated to managing individual policy assignments for users. This resource would take the following attributes:

- user_id: The ID of the Spacelift user.
- space_id: The ID of the Spacelift space.
- role: The role to assign to the user in the specified space.

### Example Usage:
```hcl
resource "spacelift_user" "john_doe" {
  username         = "john.doe"
  invitation_email = "john.doe@example.com"
}

resource "spacelift_user_policy" "john_doe_dev" {
  user_id  = spacelift_user.john_doe.id
  space_id = spacelift_space.dev.id
  role     = "WRITE"
}

resource "spacelift_user_policy" "john_doe_prod" {
  user_id  = spacelift_user.john_doe.id
  space_id = spacelift_space.prod.id
  role     = "READ"
}
```

### Benefits:

- Simplified Policy Management: Granular control over individual policy assignments without affecting other spaces.
- Reduced Risk of Errors: Minimize accidental modification of existing policy assignments.
- Increased Flexibility: Easily add or remove users from spaces without modifying their roles in other spaces.
- Improved Modularity: Separate user management from policy assignments for better organization and maintainability.

### Additional Considerations:

- The `spacelift_user` resource could retain the existing policy block for backward compatibility or bulk policy management.
[...]

PR #577, opened 2024-10-28
fix: convert User Policies to unordered set to prevent policy thrashing

bushong1:
## Description of the change

I’ve been having some trouble with the Spacelift Terraform provider when managing user access policies. Every time I add a new policy, Terraform tries to delete and recreate some of the existing ones—even though they haven’t changed at all. It’s like it’s getting confused about what’s actually new and what’s already there.

I think the problem is with how the policy field is set up in the spacelift_user resource. Right now, it’s defined as a TypeList, which is ordered. So when I insert a new policy somewhere other than the end of the list, it shifts the order, and Terraform thinks the policies have changed.

To fix this, I think we should change policy from a TypeList to a TypeSet. Sets are unordered, so Terraform won’t care about the order of the policies anymore. Also, by adding a custom hash function that combines space_id and role, we can uniquely identify each policy.

## Type of change

## Related issues

> Relates to #549, potentially sets the state to support a policy attachment feature in the future 

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/docs/resources/user.md
+++ b/docs/resources/user.md
@@ -17,7 +17,7 @@ description: |-
 
 ### Required
 
-- `policy` (Block List, Min: 1) (see [below for nested schema](#nestedblock--policy))
+- `policy` (Block Set, Min: 1) (see [below for nested schema](#nestedblock--policy))
 - `username` (String) Username of the user
 
 ### Optional
--- a/spacelift/resource_user.go
+++ b/spacelift/resource_user.go
@@ -35,7 +35,7 @@ func resourceUser() *schema.Resource {
 				Description: "Username of the user",
 			},
 			"policy": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				MinItems: 1,
 				Required: true,
 				Elem: &schema.Resource{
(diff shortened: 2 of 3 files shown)


peterdeme (approved):
it looks good to me. my only concern is that this type change (List -> Set) will show up as a change in the plan for all existing customers. could be worth to mention in the release changelog.

bushong1:
@peterdeme is the changelog something I need to do?  Also is there a way to get the "Test the code" job triggered to conclude the PR requirements?

peterdeme:
@bushong1

- Changelog: that's autogenerated on release. We (Spacelift) can update it afterwards though.
- CI/CD: the reason it won't kick off is because it contains secrets and the PR was opened by an external contributor. GitHub doesn't allow that for security reasons. We'll (Spacelift) need to open another PR copying your commit. We can do that, I'd just like to wait for @adamconnelly's review.

peterdeme:
@bushong1 thanks for the contribution! 🙏 

I ported your change into https://github.com/spacelift-io/terraform-provider-spacelift/pull/587 so that the tests can run.

PR #583, opened 2024-11-13
fix: force new when the csr changes

Apollorion:
It looks like theres no way to update the CSR after the workerpool has been created. We only set this value on create. We should force new resources if this value was to change. Its only comparing the CSR to state, so if someone was to update this today they'd think they made a change when they really didnt.


eliecharra (commented):
Good catch, thanks @Apollorion
Can you also update `spacelift/resource_worker_pool_test.go` to cover this new behavior with a test?

eliecharra (approved):

peterdeme (approved):

PR #599, opened 2025-01-23
ci: gate fork PR tests behind environment approval

eliecharra:
Run the test workflow for fork pull requests through
pull_request_target, with each run held by the preprod-tests environment
until one of its required reviewers approves it. Both jobs need preprod
secrets, which a push to a fork can never carry, so fork pull requests
produced no run at all and their required checks waited forever.

Every push to the pull request queues a fresh run that waits for its own
approval, so nothing runs unreviewed and approvals are recorded against
the exact commit. Same-repo pull requests are excluded from the gated
path, since their branches already run through the push event. The
checked out commit is pinned to the reviewed SHA, the workflow token is
read-only, and the checkout opt-in documents that fork code runs only
behind the approval gate.

## Type of change

### Development

- [x] Lint rules pass locally

### Code review


.github/workflows/test.yml
```diff
@@ -12,8 +18,27 @@ jobs:
...
+        id: checkAccess
+        uses: actions-cool/check-user-permission@7b90a27f92f3961b368376107661682c441f6103 # v2.3.0
+        with:
+          require: write
+          username: ${{ github.triggering_actor }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Check permission
+        if: steps.checkAccess.outputs.require-result == 'false' && github.actor != 'dependabot[bot]'
```
eliecharra, line 31: Allows dependabot to pass anyway. For anyone else, check if we have write access from the previous step, otherwise make the workflow fail with a few debug info about why.

peterdeme (dismissed):

eliecharra:
ping @spacelift-io/security ☝🏻

.github/workflows/test.yml
Apollorion: Cant they just remove this check with pull_request_target and run it anyways?

.github/workflows/test.yml
```diff
@@ -12,8 +18,27 @@ jobs:
...
+        run: |
+          echo "${{ github.triggering_actor }} does not have permissions on this repo."
+          echo "Current permission level is ${{ steps.checkAccess.outputs.user-permission }}"
+          echo "Job originally triggered by ${{ github.actor }}"
+          exit 1
+
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
lorengordon, line 41: I was a little curious about this particular change... Typically in a PR you would want to checkout the merge commit, rather than the head. Using the pr head will test only what is on the head branch, and will not incorporate commits that have already been merged to the base branch. Considering the other risk mitigations here, using `pull_request_target` and testing the actor permissions, do you mean to use the merge commit?

```suggestion
          ref: ${{ github.event.pull_request.merge_commit_sha }}
```

mjaworski2 (approved):
approved

Issue #603, opened 2025-02-05
Add `description` field to `spacelift_mounted_file` resource


datfinesoul:
Mounted files in Spacelift allow `Description` to be set in the Console, but it's not possible to do this via Terraform.  Could you add the ability to set `description` to the `spacelift_mounted_file` resource?

![Image](https://github.com/user-attachments/assets/643711c6-0987-4531-85e9-0a00866e65cc)

peterdeme:
https://github.com/spacelift-io/terraform-provider-spacelift/pull/604

peterdeme:
@datfinesoul thanks for noticing this. Released in v1.20.0 🙇

PR #604, opened 2025-02-06
Add description to `spacelift_mounted_file`

peterdeme:
## Description of the change

Resolving https://github.com/spacelift-io/terraform-provider-spacelift/issues/603

## Type of change

### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/spacelift/resource_mounted_file.go
+++ b/spacelift/resource_mounted_file.go
@@ -73,6 +73,12 @@ func resourceMountedFile() *schema.Resource {
 				Optional:    true,
 				ForceNew:    true,
 			},
+			"description": {
+				Type:        schema.TypeString,
+				Description: "Free-form description of the mounted file",
+				Optional:    true,
+				ForceNew:    true,
+			},
 			"write_only": {
 				Type:        schema.TypeBool,
 				Description: "Indicates whether the content can be read back outside a Run. Defaults to `true`.",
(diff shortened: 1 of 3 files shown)


spacelift/resource_mounted_file.go
```diff
@@ -73,6 +73,12 @@ func resourceMountedFile() *schema.Resource {
 				Optional:    true,
 				ForceNew:    true,
 			},
+			"description": {
+				Type:        schema.TypeString,
+				Description: "Free-form description of the mounted file",
+				Optional:    true,
+				ForceNew:    true,
```
eliecharra, line 80: Should this really trigger resource recreate?
```suggestion
				ForceNew:    false,
```

peterdeme: yes, it should. we don't have an "update resource" method implemented here, and I would need to implement one just for this. the implementation of it isn't super obvious.


Also, this follows the same pattern as `resource_environment_variable`. it's not ideal, but it'll work.

https://github.com/spacelift-io/terraform-provider-spacelift/blob/dee0f609998116bc85863c84c7d68bc88015e5a7/spacelift/resource_environment_variable.go#L85-L90

eliecharra: Got it, thanks for the context. I just wanted to be sure it was not a copy pasta error 🙏🏻

eliecharra (approved):

spacelift/resource_mounted_file.go
peterdeme: thanks for noticing anyway!

PR #652, opened 2025-07-14
feat: add ability for stack destructor to discard runs

Apollorion:


docs/resources/stack_destructor.md
```diff
@@ -39,6 +39,7 @@ resource "spacelift_stack_destructor" "k8s-core" {
 
 ### Optional
 
+- `cancel_pending_runs` (Boolean) If set to true, destruction will also cancel any pending runs
```
michieldewilde, line 42: Could you elaborate a bit more on `cancel any pending runs`. Cancel is quite a loaded term in Spacelift and this does not only cancel pending runs in the strict sense of the term.

michieldewilde (dismissed):

docs/resources/stack_destructor.md
Apollorion: Can you elaborate on what you're thinking? In my head, a cancelled run is exactly what it sounds like. Any run that can possibly be discarded will be.

PR #663, opened 2025-08-11
Git checkout option

Axot017:
## Description of the change

Add `use_git_checkout` option to each integration

## Type of change

## Related issues

> Fix [#1]()

### Development

- [ ] Lint rules pass locally

### Code review


docs/resources/bitbucket_datacenter_integration.md
```diff
@@ -54,6 +54,7 @@ resource "spacelift_bitbucket_datacenter_integration" "private-example" {
 - `description` (String) Bitbucket Datacenter integration description
 - `labels` (Set of String) Bitbucket Datacenter integration labels
 - `space_id` (String) Bitbucket Datacenter integration space id. Defaults to `root`.
+- `use_git_checkout` (Boolean) Indicates whether the integration should use git checkout. If false source code will be downloaded using the VCS API. Defaults to false.
```
mbialon, line 57: The default is `true` for some resources and `false` for others. Is that to be in line with the defaults in SaaS?

spacelift/type_conversions.go
```diff
@@ -88,3 +88,19 @@ func toOptionalStringMap(input interface{}) *structs.StringMap {
 	}
 	return &customHeaders
 }
+
+// getOptionalBool checks if user has explicitly set a boolean value in the configuration and returns it as a pointer to graphql.Boolean.
+// If the value is not set in the configuration, it returns nil.
+func getOptionalBool(d *schema.ResourceData, key string) *graphql.Boolean {
```
mbialon, line 94: Could we provide some tests?

spacelift/type_conversions.go
```diff
@@ -88,3 +88,19 @@ func toOptionalStringMap(input interface{}) *structs.StringMap {
...
+	// Check if the raw config has the key set explicitly
+	rawConfig := d.GetRawConfig()
+	if !rawConfig.IsNull() {
```
mbialon, line 97: Nit: I prefer to keep the code as close to the left as possible. I believe it makes the code path easier to read.
Could we invert the conditions and go with early returns?

spacelift/resource_gitlab_integration_test.go
```diff
@@ -72,7 +74,7 @@ func TestGitLabIntegrationResource(t *testing.T) {
 
 		testSteps(t, []resource.TestStep{
 			{
-				Config: configGitLab(host, token, descr, "null", "null"),
+				Config: configGitLab(host, token, descr, "null", "null", false),
```
mbialon, line 77: Nit: If we defined a const `useGitCheckout` at the top of the tests and used it here like `!useGitCheckout`, it would make what that bool means way more readable.

docs/resources/bitbucket_datacenter_integration.md
Axot017: >  Is that to be in line with the defaults in SaaS?

Exactly

spacelift/type_conversions.go
Axot017: I tried to make some unit tests for this function and failed. `*schema.ResourceData` contains only private fields and test util function provided by terraform to generate `*schema.ResourceData` does not populate fields that `GetRawConfig` is using.

Technically this function has some coverage, because its invoked by resources that are tested

tomasmik (approved):

eliecharra (approved):

PR #664, opened 2025-08-13
docs: clarify spacelift_stack_destructor `deactivated` attribute

oycyc:
The current description of `If set to true, destruction won't delete the stack` is not really accurate, because what it actually does is preserve the resources taht the stack manages, as the description correctly specify. This confused me when reading the docs
image

### DIFF
--- a/docs/resources/stack_destructor.md
+++ b/docs/resources/stack_destructor.md
@@ -39,7 +39,7 @@ resource "spacelift_stack_destructor" "k8s-core" {
 
 ### Optional
 
-- `deactivated` (Boolean) If set to true, destruction won't delete the stack
+- `deactivated` (Boolean) If set to true, destruction won't delete the resources (e.g. AWS/Azure/GCP infrastructure) that the stack manages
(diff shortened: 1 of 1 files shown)


docs/resources/stack_destructor.md
```diff
@@ -39,7 +39,7 @@ resource "spacelift_stack_destructor" "k8s-core" {
 
 ### Optional
 
-- `deactivated` (Boolean) If set to true, destruction won't delete the stack
+- `deactivated` (Boolean) If set to true, destruction won't delete the resources (e.g. AWS/Azure/GCP infrastructure) that the stack manages
```
oycyc, line 42:
```suggestion
- `deactivated` (Boolean) If set to true, destruction won't delete the stack & the resources (e.g. AWS/Azure/GCP infrastructure) that the stack manages
```

peterdeme:
@oycyc the docs are auto-generated from the resources. can you please update this line?

https://github.com/spacelift-io/terraform-provider-spacelift/blob/15a6902305d5488a621f33084346b151c6426e03/spacelift/resource_stack_destructor.go#L49

and then `cd tools && go generate ./...`

peterdeme (approved):

PR #691, opened 2025-10-21
Remove sign flag from goreleaser command when building local provider

michieldewilde:
## Description of the change

https://goreleaser.com/deprecations/#-skip removed in v2.0

## Type of change

- [x] Documentation (non-breaking change that adds documentation)


peterdeme (approved):
yeah I noticed this myself yesterday, but I was too lazy to make this change :D

PR #776, opened 2026-04-10
feat(spacelift_run): add runtime_config support for custom per-run configuration

### LINKED ISSUE #775: `spacelift_run`: Support custom runtime configuration (environment variables and TF var overrides)
## Summary

The `spacelift_run` resource currently supports triggering runs on a stack, but has no way to pass custom runtime configuration, environment variables or Terraform variable overrides scoped to a single run.

This capability already exists in the Spacelift UI ("Trigger with custom runtime config") and the GraphQL API, but the Terraform provider doesn't expose it.

## Use Case

When programmatically triggering runs, it's common to need per-run configuration that differs from the stack's default. For example:

- Passing a one-time variable override to control behavior for a specific run
- Setting an environment variable that should not persist on the stack
- Triggering a targeted plan/apply with runtime-specific parameters

## Proposed Change

Add an optional block to `spacelift_run` for runtime configuration:

```hcl
resource "spacelift_run" "example" {
  stack_id = spacelift_stack.this.id

  runtime_config {
    environment_variable {
      key   = "TARGET_RESOURCE"
      value = "aws_instance.example"
    }

    terraform_variable {
      key   = "enable_migration"
      value = "true"
    }
  }
}
```

Azmah-Bad:
## Summary

- Switch `spacelift_run` from `runResourceCreate` to `runTrigger` GraphQL mutation, enabling custom runtime configuration (environment variables, project root, runner image, and before/after hook scripts) scoped to a single run
- Matches the "Trigger with custom runtime config" capability already available in the Spacelift UI
- Deduplicate runtime config parsing by extracting shared `parseRuntimeConfig()` used by both `spacelift_run` and `spacelift_scheduled_run`
- The existing `proposed` field is preserved for backward compatibility, mapped internally to the `runType` enum

Tested locally with a dev_overrides provider build against a live Spacelift stack — confirmed runtime config environment variables are passed through and visible in the UI and GraphQL API.

Closes #775

---

> [!NOTE]
> **Medium Risk**
> Changes the GraphQL mutation used to trigger runs and adds new runtime configuration inputs, which could affect run creation behavior and backward compatibility if the API differs. Scope is limited to run-triggering resources and includes test coverage and docs updates.
> 
> **Overview**
> **`spacelift_run` can now trigger runs with per-run `runtime_config`.** The resource schema and docs add a `runtime_config` block supporting env vars, `project_root`, `runner_image`, and before/after hook script lists.
> 
> Run creation is switched from the legacy `runResourceCreate` mutation to `runTrigger`, mapping the existing `proposed` boolean to a new `RunType` enum (`PROPOSED`/`TRACKED`) while preserving the old field. Runtime-config parsing is deduplicated into a shared `parseRuntimeConfig()` reused by `spacelift_scheduled_run`, and acceptance tests add coverage for the new `runtime_config` block.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4929f4c913b221bdb8f0559a2062de0f46578c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/resources/run.md
+++ b/docs/resources/run.md
@@ -40,12 +40,43 @@ resource "spacelift_run" "this" {
 - `commit_sha` (String) The commit SHA for which to trigger a run.
 - `keepers` (Map of String) Arbitrary map of values that, when changed, will trigger recreation of the resource.
 - `proposed` (Boolean) Whether the run is a proposed run. Defaults to `false`.
+- `runtime_config` (Block List, Max: 1) Custom runtime configuration for this run. (see [below for nested schema](#nestedblock--runtime_config))
 - `timeouts` (Block, Optional) (see [below for nested schema](#nestedblock--timeouts))
 - `wait` (Block List, Max: 1) Wait for the run to finish (see [below for nested schema](#nestedblock--wait))
 
 ### Read-Only
 
 - `id` (String) The ID of the triggered run.
+
+<a id="nestedblock--runtime_config"></a>
+### Nested Schema for `runtime_config`
+
+Optional:
+
+- `after_apply` (List of String) List of after-apply scripts.
+- `after_destroy` (List of String) List of after-destroy scripts.
+- `after_init` (List of String) List of after-init scripts.
+- `after_perform` (List of String) List of after-perform scripts.
+- `after_plan` (List of String) List of after-plan scripts.
+- `after_run` (List of String) List of after-run scripts.
+- `before_apply` (List of String) List of before-apply scripts.
+- `before_destroy` (List of String) List of before-destroy scripts.
+- `before_init` (List of String) List of before-init scripts.
+- `before_perform` (List of String) List of before-perform scripts.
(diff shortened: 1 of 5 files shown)


sephriot (approved):

Azmah-Bad:
Hi @sephriot, thanks for the review! 

I tested this PR locally and it works, however when running it on spacelift's runner I get an error because I can't access the `runTrigger` mutation from inside a spacelift public runner. Is this a known limitation?

peterdeme:
@Azmah-Bad I think we'll rather extend the existing `RunResourceCreate` API with runtime options instead of exposing the RunTrigger method.

michieldewilde:
We are closing this PR because we will be making the change to use the run trigger API instead of the run resource create API for this resource. related to #796

PR #796, opened 2026-05-21
feat(resource_run): switch to run trigger API

michieldewilde:
## Description of the change

Swap API from run resource create to run trigger.

Adds a `runtime_config` nested block to `spacelift_run` so users can override the stack's runtime configuration when triggering a run (runner image, project root, environment variables, before/after hook scripts). The underlying `runTrigger` GraphQL mutation already accepts a `runtimeConfig` input; this wires it through the provider.

All `runtime_config` fields on `spacelift_run` are marked `ForceNew` because the resource is immutable; changing the config should trigger a new run rather than silently no-op.

## Type of change

## Related issues

> N/A

### Development

- [x] Lint rules pass locally
- [x] Default values have been documented in the description

### Code review

### DIFF
--- a/spacelift/resource_run.go
+++ b/spacelift/resource_run.go
@@ -106,28 +106,25 @@ func resourceRun() *schema.Resource {
 
 func resourceRunCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	var mutation struct {
-		ID string `graphql:"runTrigger(stack: $stack, commitSha: $sha, runType: $runType, runtimeConfig: $runtimeConfig)"`
+		CreateRun structs.Run `graphql:"runTrigger(stack: $stack, commitSha: $sha, runType: $runType, runtimeConfig: $runtimeConfig)"`
 	}
 
 	stackID := d.Get("stack_id").(string)
 
+	runType := structs.RunTypeTracked
+	if d.Get("proposed").(bool) {
+		runType = structs.RunTypeProposed
+	}
+
 	variables := map[string]any{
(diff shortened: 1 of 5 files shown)


peterdeme (dismissed):

peterdeme (approved):

marko-kos (approved):

spacelift/resource_run.go
```diff
@@ -92,43 +92,57 @@ func resourceRun() *schema.Resource {
...
 func resourceRunCreate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
 	var mutation struct {
-		ID string `graphql:"runResourceCreate(stack: $stack, commitSha: $sha, proposed: $proposed)"`
+		CreateRun structs.Run `graphql:"runTrigger(stack: $stack, commitSha: $sha, runType: $runType, runtimeConfig: $runtimeConfig)"`
 	}
 
 	stackID := d.Get("stack_id").(string)
 
+	runType := structs.RunTypeTracked
+	if d.Get("proposed").(bool) {
```
aerfio, line 115: Other type assertions in this fn seem to rely on the 2-value test assertion
```go
val, ok := variable.(type)
```

aerfio (approved):
Apart from this one nitpicky thing LGTM, approving already to not block

spacelift/resource_run.go
michieldewilde: It's a `TypeBool` field, so `d.Get` is guaranteed to return a `bool` - no need for the 2-value form here.

PR #797, opened 2026-05-21
feat(space): add adopt_existing flag to spacelift_space resource

KNiepok:
## Summary

Adds an `adopt_existing` boolean to the `spacelift_space` resource. When set:

- **Create**: calls the new backend `spaceCreateOrGet` mutation. If a space with the same `name` already exists under the same `parent_space_id`, the existing row is adopted into Terraform state; otherwise a new one is created.
- **Update**: unchanged. Goes through the normal `spaceUpdate` mutation — last writer wins across multiple adopters.
- **Delete**: state-only no-op. The backend space is **not** removed. See "Destroy semantics" below.

Default `false` — existing configs are unchanged.

Motivation: shared Terraform modules instantiated by many admin stacks need to declare the same logical space without erroring or producing silent duplicate rows. Customer: Trainline.

**Blocked on backend PR: https://github.com/spacelift-io/backend/pull/14293** — that one must merge + deploy to prod before this PR is taken out of draft and a provider release is tagged.

ClickUp: https://app.clickup.com/t/869dbya74

## Design decisions

- **Opt-in flag** (default `false`) rather than changing `spaceCreate` semantics. Reversible if drift behaviour bites; doesn't shift semantics for any other caller (UI, spacectl, third-party integrations).
- **Adopt path ignores `description` / `inherit_entities` / `labels`** on the create-time call. Subsequent Updates use the regular `spaceUpdate` mutation — standard last-writer-wins. Docs warn callers to compute identical attributes from shared inputs to avoid drift wars between competing stacks.
- **Name match on the backend is case-sensitive.**
- **Adopt-time authz** (enforced on the backend): the caller must have admin access on the existing space (or be a root admin). Parent-admin alone is not sufficient.

## Destroy semantics

`terraform destroy` on a resource with `adopt_existing = true` removes the resource from Terraform state without calling `spaceDelete`. With many stacks adopting the same space, a single destroy from one stack would otherwise either:

1. Fail with `ErrSpaceHasDependants` (other stacks still have entities attached to the space), leaving that stack's state stuck, or
2. Succeed and silently strand every other adopter's state pointing at a deleted backend ID, triggering recreation on their next `apply`.

Disown-on-destroy is the safe default for adopt-style resources (mirrors AWS provider's `lifecycle { prevent_destroy = true }` pattern for shared-ownership resources). To actually delete the space: take ownership in one configuration with `adopt_existing = false` and `terraform destroy` from there, or remove it via the UI / spacectl.

## Files

- `spacelift/resource_space.go` — `adopt_existing` schema field, branched `resourceSpaceCreate`, branched `resourceSpaceDelete`
- `spacelift/internal/structs/space.go` — new `SpaceCreateOrGetPayload` struct
- `spacelift/resource_space_test.go` — acceptance test for adopt path (two resources with same name end up with the same ID)
- `examples/resources/spacelift_space/resource.tf` — shared-module example
- `docs/resources/space.md` — manually updated to mirror `tfplugindocs` output
- `.gitignore` — `.worktrees/`

## Test plan

- [ ] Backend PR merged + deployed to prod
- [ ] `TF_ACC=1 go test -v --timeout 200s -race -run TestSpaceResource ./spacelift/` (acceptance test against staging that has the new mutation)
- [ ] `go generate ./tools/...` and confirm `docs/resources/space.md` matches the manual update
- [ ] `golangci-lint run ./...`
- [ ] Manual: two `spacelift_space` resource blocks with same name+parent and `adopt_existing = true` in two stacks → both apply cleanly, both end up with the same `id`
- [ ] Manual: `terraform destroy` on one of the adopters → backend space survives, other adopter's `terraform plan` shows no diff
- [ ] Manual: drop `adopt_existing = true` (or leave default `false`) and apply two stacks with same name → second errors as today

## Notes for reviewer

- The doc was updated by hand to avoid churn. Running `go generate ./tools/...` should be idempotent and produce the same output; if not, that's a CI signal to investigate.
- The provider release tag is intentionally not part of this PR — release only after the backend has shipped the mutation.


Clickup-user:
Task linked: [CU-869dbya74 Add `adopt_existing` to `spacelift_space` resource (backend `spaceCreateOrGet` + provider flag)](https://app.clickup.com/t/869dbya74)

PR #822, opened 2026-07-22
fix(bitbucket-cloud): support API token auth in integration data source

Yantrio:
## Description of the change

Bitbucket Cloud integrations authenticated with an **Atlassian API token** (email + token) leave the `username` field **empty** server-side.

The username field should now be deprecated as it's no longer supported by bitbucket cloud.

### Changes

- **Add `email`** attribute to the `spacelift_bitbucket_cloud_integration` data source (read-only)
- **Deprecate `username`** — it's only populated for legacy app-password integrations.

## Type of change

- [ ] Chore
- [ ] Breaking change
- [ ] Documentation

🤖 Generated with [Claude Code](https://claude.com/claude-code)


michalrom089:
@Yantrio I was wondering about deprecating this field in our GraphQL schema, and realized that we don't allow creating `BitbucketCloudIntegration` here, right? I can only see `dataBitbucketCloudIntegration`

Yantrio:
>@Yantrio I was wondering about deprecating this field in our GraphQL schema, and realized that we don't allow creating `BitbucketCloudIntegration` here, right? I can only see `dataBitbucketCloudIntegration` 

Correct, we don't allow setting up of cloud integrations or vcs integrations in the provider if I remember correctly, because most of them require human intevention or setup.

It's purely data sources to fetch the info.

tomasmik (approved):

PR #827, opened 2026-07-28
feat(repos): add the spacelift_repo_file resource

sephriot:
## Description of the change

A repo on its own has no content and cannot be attached to a stack. This adds files, so a repo becomes usable.

Files are written as commits, and a commit happens only when the contents, permissions or encryption of a file change. The commit message and author describe the next such commit rather than causing one, so correcting a message does not touch the repo or start runs on stacks using it. Leaving a file out of a commit keeps it as it was; removing it takes an explicit delete.

> Spacelift never returns the contents of an encrypted file, so Terraform cannot tell when one has been changed elsewhere.

## Type of change

## Related issues

> Fix [#1]()

### Development

- [ ] Lint rules pass locally

### Code review


spacelift/resource_repo_file.go
```diff
@@ -0,0 +1,263 @@
...
+				Required:         true,
+				ForceNew:         true,
+				ValidateDiagFunc: validations.DisallowEmptyString,
+			},
+			repoFileContent: {
+				Type:        schema.TypeString,
+				Description: "Contents of the file. Text only - binary files are not supported.",
+				Required:    true,
+			},
+			repoFileMode: {
```
kutluhanmetin, line 72: `file_mode = "644"` causes a permanent plan diff: the backend accepts any octal but always returns `"0644"`, so after every apply the next plan still shows `~ file_mode = "0644" -> "644"`.

The solution would be to add:
```
DiffSuppressFunc: func(_, old, new string, _ *schema.ResourceData) bool {
	o, err1 := strconv.ParseUint(old, 8, 32)
	n, err2 := strconv.ParseUint(new, 8, 32)
	return err1 == nil && err2 == nil && o == n
}, 
```

sephriot: done

kutluhanmetin (dismissed):

kutluhanmetin (approved):

PR #830, opened 2026-07-28
test: make acceptance tests genuinely parallel

michieldewilde:
## Description of the change

The test job took ~23 minutes and spent nearly all of it waiting rather than working. `-parallel 40` was barely being used.

`t.Parallel()` was only ever called inside the `testSteps` helper, on whatever `*testing.T` it was handed, so parallelism depended on *where* the helper was called from:

| Shape | Count | Before |
|---|---|---|
| `testSteps(t, …)` directly on the top-level `t` | 29 | already parallel |
| `testSteps` only inside `t.Run` closures | 122 | **sequential** — a parallel subtest resumes only after its parent returns |

That left ~122 sequential groups plus a concurrent tail of 29, i.e. effective concurrency of ~2.8 against a budget of 40. This adds `t.Parallel()` to the 122-function group only; adding it to the other 29 panics with `t.Parallel called multiple times`.

Two things had to change alongside it.

**The shared provider instance.** `testProvider()` memoised one `*schema.Provider` for the whole suite. `Providers` is deprecated in favour of `ProviderFactories` for exactly this reason: Terraform mutates the provider when it configures it, and the SDK serves it in-process, so every concurrent test hit the same object. `go test -race` reports 2 data races on `grpc_provider.go:735` with the singleton and 0 with a factory. `test_provider_framework.go` already did this correctly with `ProtoV6ProviderFactories`; the SDKv2 helper never caught up.

**Account-wide settings.** `Test_resourceDefaultAccountRunnerImage` points the account's default runner image at tags that exist nowhere, and any stack that doesn't set its own `runner_image` inherits it — so once everything ran at once, unrelated tests' workers started dying with `pull access denied for public-runner`. All 13 run failures in one window carried that byte-identical note. It's now sequential, as is `Test_resourceSecurityEmail`, which writes account-level state the same way; `Test_resourceAuditTrailWebhook` already was. Note that test was *already* parallel before this PR — it just used to overlap 28 others rather than 150, so collisions looked like background noise.

### Result

| | Before | After |
|---|---|---|
| `./spacelift` package | 1375.7s | ~210s |
| Whole job | 23m25s | ~4m |

## Type of change

- [x] Chore (maintenance work, dependency bumps, refactors, not supposed to break existing functionalities)

### Code review

- [x] This pull request has a descriptive title and information useful to a reviewer

### DIFF
--- a/spacelift/data_aws_integration_attachment_external_id_test.go
+++ b/spacelift/data_aws_integration_attachment_external_id_test.go
@@ -12,6 +12,8 @@ import (
 )
 
 func TestAWSIntegrationAttachmentExternalIDData(t *testing.T) {
+	t.Parallel()
+
 	const resourceName = "data.spacelift_aws_integration_attachment_external_id.test"
 
 	t.Run("with a stack", func(t *testing.T) {
--- a/spacelift/data_aws_integration_attachment_test.go
+++ b/spacelift/data_aws_integration_attachment_test.go
@@ -11,6 +11,8 @@ import (
 )
 
 func TestAWSIntegrationAttachmentData(t *testing.T) {
+	t.Parallel()
+
 	const resourceName = "data.spacelift_aws_integration_attachment.test"
 
 	t.Run("with a stack", func(t *testing.T) {
--- a/spacelift/data_aws_integration_test.go
+++ b/spacelift/data_aws_integration_test.go
@@ -12,6 +12,8 @@ import (
 )
 
 func TestAWSIntegrationData(t *testing.T) {
+	t.Parallel()
+
 	t.Run("without generating AWS creds in the worker", func(t *testing.T) {
 		randomID := acctest.RandStringFromCharSet(5, acctest.CharSetAlphaNum)
 		testSteps(t, []resource.TestStep{{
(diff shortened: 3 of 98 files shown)


sephriot (approved):
Very promising

aerfio:
Could you add [tparallel](https://golangci-lint.run/docs/linters/configuration/#tparallel) linter to golangci config? Or at least add it locally and run it, if we dont want to add too many linters there? 

https://golangci-lint.run/docs/linters/configuration/#paralleltest also might ensure that new tests set this flag in the future, up to you whether you actually add it, I'm just mentioning it.

michieldewilde:
> Could you add [tparallel](https://golangci-lint.run/docs/linters/configuration/#tparallel) linter to golangci config? Or at least add it locally and run it, if we dont want to add too many linters there?
> 
> https://golangci-lint.run/docs/linters/configuration/#paralleltest also might ensure that new tests set this flag in the future, up to you whether you actually add it, I'm just mentioning it.

addressed in https://github.com/spacelift-io/terraform-provider-spacelift/pull/831
For paralleltest: it requires a literal t.Parallel() in the function body and can't see through the testSteps helper.

peterdeme (approved):

PR #843, opened 2026-08-29
feat: add spacelift_ai_integration resource and data sources

derhally:
## Description of the change

Adds a `spacelift_ai_integration` resource with full CRUD and import, plus `spacelift_ai_integration` and `spacelift_ai_integrations` data sources. The `aiIntegration` API had no Terraform coverage.

All four providers are supported. Since the API models AI integrations as a single object discriminated by a `provider` field, this is one resource rather than one per provider. The provider is chosen by setting exactly one provider block, the way `spacelift_stack` expresses its vendor:

```hcl
resource "spacelift_ai_integration" "gemini" {
  name = "gemini"

  google {
    api_key_wo         = var.gemini_api_key
    api_key_wo_version = "1"
  }
}
```

Four API behaviours drove the design:

- `enabled` is not accepted by `aiIntegrationCreate`/`Update`, so it is reconciled through `aiIntegrationToggle`. This also applies to Spacelift-provided integrations, which are otherwise read-only.
- The API never returns the key, so it follows the write-only pattern from `spacelift_named_webhook`. `ExtractWriteOnlyField` gained a nested variant for fields inside a `MaxItems: 1` block. The top-level path is unchanged.
- `models` is rejected outright for Bedrock, which derives them from inference profiles instead, so the schema makes that a plan-time conflict.
- `aiIntegrationUpdate` accepts a `provider` argument but the API rejects changing it, so swapping the provider block replaces the integration.

AI was used to implement this feature.

## Type of change

## Related issues

None.

## Testing notes

The Bedrock tests need a real AWS integration and inference profile, so they skip unless `SPACELIFT_PROVIDER_TEST_AI_BEDROCK_INTEGRATIONID` and `_PROFILE` are set. The rest run against a placeholder key, which the API accepts. All were verified against a live account.


spacelift/resource_ai_integration.go
```diff
@@ -0,0 +1,577 @@
...
+	}
+
+	return d.Set(block, []any{entry})
+}
+
+func resourceAIIntegrationUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	// Spacelift-provided integrations are read-only, but aiIntegrationToggle
+	// explicitly works on them, so only reject changes to anything else.
+	spaceliftProvided := d.Get("is_spacelift_provided").(bool)
+	if spaceliftProvided && d.HasChangesExcept("enabled") {
```
rekfuki, line 628: I would suggest adding a fifth provider block, `spacelift`, for the Spacelift-provided integrations, because otherwise managing one would require a placeholder credential, and with `api_key_wo` it doesn't work. Import would leave `api_key_wo_version` empty in state while the config supplies `"1"`. That diff hits the read-only check before the toggle runs, the failed apply never persists the version, and the same plan repeats forever.

The config could look something like this:

```hcl
data "spacelift_ai_integrations" "shared_anthropic" {
  spacelift_provided = true   # see the data source comment
  ai_provider        = "Anthropic"
}

import {
  to = spacelift_ai_integration.shared_anthropic
  id = one(data.spacelift_ai_integrations.shared_anthropic.integrations[*].integration_id)
}

resource "spacelift_ai_integration" "shared_anthropic" {
  enabled = false

  spacelift {}   # marks a Spacelift-provided integration, nothing to configure
}
```

spacelift/data_ai_integrations.go
```diff
@@ -0,0 +1,171 @@
...
+func dataAIIntegrations() *schema.Resource {
+	return &schema.Resource{
+		Description: "" +
+			"`spacelift_ai_integrations` represents a list of all the AI integrations in the Spacelift " +
+			"account visible to the API user, including the read-only ones provided by Spacelift.",
+
+		ReadContext: dataAIIntegrationsRead,
+
+		Schema: map[string]*schema.Schema{
+			"labels": {
```
rekfuki, line 23: Could you add boolean `spacelift_provided` and string `ai_provider` filters alongside `labels`? Otherwise finding the Spacelift-provided integration to import would take a for comprehension over the full list. Client-side filtering like `labels` does is fine, the list is small. Note that `spacelift_provided` needs a raw config null check, since `d.Get` can't tell unset from false.

spacelift/internal/helpers.go
```diff
@@ -9,18 +9,44 @@ import (
...
+) (string, diag.Diagnostics) {
 	var result string
 
-	if v, ok := data.GetOk(value); ok {
+	if v, ok := data.GetOk(key(value)); ok {
 		result = v.(string)
 	}
 
-	if _, ok := data.GetOk(valueWoVersion); ok {
+	if _, ok := data.GetOk(key(valueWoVersion)); ok {
```
rekfuki, line 45: The API key should only be extracted and sent when creating or when `api_key_wo_version` actually changed. Otherwise every update would resend the key and the API treats that as a rotation. With a custom `base_url` it re-verifies the gateway and probes the models, so a description-only apply can fail on gateway downtime or an out-of-band key revocation.

spacelift/resource_ai_integration.go
```diff
@@ -0,0 +1,577 @@
...
+						},
+						"integration_name": {
+							Type:        schema.TypeString,
+							Description: "Name of the AWS integration used to reach Bedrock",
+							Computed:    true,
+						},
+					},
+				},
+			},
+			"models": {
```
rekfuki, line 143: The provider should send an actual empty list when `models = []` is explicitly set in config. Otherwise that config would plan a change but the provider sends `null`, the API keeps the stored list, and the same diff comes back on every plan. The API does support clearing, an explicit empty list resets to the default model list, so this would also make un-pinning possible instead of one-way.

The same check also stops the Computed default list being re-sent and stored on the first unrelated update, which today silently pins the integration to a snapshot of the defaults.

spacelift/resource_ai_integration.go
```diff
@@ -0,0 +1,577 @@
...
+		Description: fmt.Sprintf(
+			"%s-specific configuration. Presence means this integration uses the %s API.",
+			label, api,
+		),
+		Optional:     true,
+		MaxItems:     1,
+		ExactlyOneOf: aiProviderBlockNames,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"api_key": {
```
rekfuki, line 223: `api_key` and `api_key_wo` should get `DisallowEmptyString` like the other string fields in the resource, because otherwise an empty string would satisfy `AtLeastOneOf` and the misconfiguration (typically an unset variable in CI) only surfaces as an API error mid-apply instead of at plan time.

rekfuki (changes requested):

derhally:
@rekfuki I believe I've addressed the issues you raised.

On `models = []`: you were right that the API clears the pin, but I could not make it settle.

Reproduced against a live account: pinned `gemini-2.5-pro` and `gemini-2.5-flash`, then applied `models = []`. The read immediately after the update returned `[]`, but the next refresh returned `gemini-3.1-pro-preview` and `gemini-3.5-flash`. Config says `[]`, state says two models, so every plan wants to clear it again.

The problem is that `AIIntegration.models` is NON_NULL and reports the models the integration resolves to, not the ones it was given, so there is no way to observe that nothing is pinned. The gap between those two reads also looks like an async backfill server side, which may be worth a bug on its own.

If `models` stayed empty once explicitly cleared, or the resolved list moved to `modelDetails`, this would work exactly as you described and I would switch back straight away.

For now I have put `MinItems: 1` on `models`, so an empty list fails at plan time rather than diffing forever, and pinning is documented as one-way. Setting a non-empty list and changing it later both update in place and plan clean, verified live.

Which way do you want to go?

rekfuki:
> @rekfuki I believe I've addressed the issues you raised.
> 
> On `models = []`: you were right that the API clears the pin, but I could not make it settle.
> 
> Reproduced against a live account: pinned `gemini-2.5-pro` and `gemini-2.5-flash`, then applied `models = []`. The read immediately after the update returned `[]`, but the next refresh returned `gemini-3.1-pro-preview` and `gemini-3.5-flash`. Config says `[]`, state says two models, so every plan wants to clear it again.
> 
> The problem is that `AIIntegration.models` is NON_NULL and reports the models the integration resolves to, not the ones it was given, so there is no way to observe that nothing is pinned. The gap between those two reads also looks like an async backfill server side, which may be worth a bug on its own.
> 
> If `models` stayed empty once explicitly cleared, or the resolved list moved to `modelDetails`, this would work exactly as you described and I would switch back straight away.
> 
> For now I have put `MinItems: 1` on `models`, so an empty list fails at plan time rather than diffing forever, and pinning is documented as one-way. Setting a non-empty list and changing it later both update in place and plan clean, verified live.
> 
> Which way do you want to go?

Sorry for the slow reply @derhally, and thanks for the repro. Turned out to be two bugs sitting on top of each other.

First, models was reporting what the integration resolves to rather than what's actually pinned, so clearing the pin didn't show up anywhere. Second (and this is where your `[]` came from), when the resolver couldn't load the default catalogue it swallowed the error and just returned an empty list. So depending on whether the catalogue load happened to succeed you'd get different answers back to back.

Both fixed now. `AIIntegration.models` is nullable: `null` means nothing is pinned and the integration follows the default list, a non-null list is exactly the pinned models (can be empty). `modelDetails` still gives you the resolved models if you need them. Bedrock is unchanged, it always reports the stored profile IDs. And the resolver surfaces the catalogue error instead of hiding it, so reads should be consistent.

This means your original design should just work and you can drop the `MinItems: 1`. `models = []` sends an empty non-null list (your raw-config path already does this), the API clears the pin and reads back null, plan converges. Unset sends null, reads null, no diff, which also gets rid of the silent default pinning. Non-empty pins those models and you get the same list back.

Read side shouldn't need anything since the client decodes null to a nil slice and setting a nil slice clears state either way.

derhally:
Thanks @rekfuki, confirmed the fix is live and `MinItems: 1` is gone. 

`models = []` now clears the pin and reads back null, plan converges. Verified against a live account, along with pinning, re-pinning, and unset on create.

One thing you may want to look at: on `aiIntegrationUpdate`, `models: null` leaves the existing pin in place rather than clearing it. Probed directly against the API:

| argument | result |
| --- | --- |
| `models: null` | pin unchanged |
| `models: []` | pin cleared, reads back null |

That matters when someone pins models and later deletes the line from their config. Sending null there would leave the integration pinned to models the config no longer mentions, so the provider now sends `[]` whenever the config pins nothing. Bedrock still sends null, since the API rejects the `models` argument for it and takes the profiles through `providerConfig` instead.

Is null meaning leave-alone on update intended, or should it clear the pin like an empty list does?

rekfuki:
Yes, null meaning leave-alone on update is intended. The mutation takes every argument on every call, so null has to mean "don't touch", otherwise a description-only update would wipe the pin.

Sending `[]` for an absent attribute has a catch though, as long as `models` keeps `Computed`. With Optional plus Computed the SDK plans the prior state value when the attribute is absent from the config, so deleting the `models` line produces no plan. Your `[]` only goes out when some unrelated change triggers an update, and the unpin happens without a plan showing it. It would also clear pins made in the UI on configs that never mention `models`.

I'd drop `Computed` from `models` instead. `Computed` is what let the state accept the default list the old API filled into the read without a diff, and now that unpinned reads back null there's nothing left to compute. With plain Optional an absent attribute plans a visible clear, so deleting the pin line unpins and the plan shows it, and `models = []` does the same. A pin made in the UI on a Terraform-managed integration shows up as drift and gets reverted in a plan, exactly how `labels` already behaves in this resource. Bedrock keeps sending null since the API rejects the argument there.

This does mean absent stops meaning "not managed here" and Terraform always owns the field, so a pin someone makes in the UI gets reverted on the next apply. I'd say that's fine, when the integration is under Terraform then Terraform should be the source of truth, not clickops. `labels` and `description` already work exactly like that.

derhally:
Computed is gone. Deleting the models line now plans a visible clear and unpins, models = [] does the same, and a pin made in the UI comes back as drift and gets reverted. All checked against a live account, and the drift case has a test that pins through the API between steps.

Computed was also what allowed state to hold values the config can't set. Bedrock reports its profiles in models, but models conflicts with the bedrock block, so plain Optional turned that into a permanent diff. I fixed it by not recording models for Bedrock at all, the same way description and labels are already skipped for Spacelift-provided integrations. Practical effect: models is always empty in state for Bedrock, and you read the profiles from bedrock.profiles or the data sources.

rekfuki (approved):
@derhally Thanks for this - LGTM 🎉 

Some of the commits are not signed so you'll have to fix that otherwise we won't be able to merge

derhally:
@rekfuki commits should be signed now

rekfuki:
Merged. Thanks for contributing, I'll get this released 🎉

PR #846, opened 2026-09-03
stacks: Add `lock` attribute and optionally prevent changes when locked

zhimsel:
## Description of the change

This PR adds a `lock{}` attribute to all "stack" resource schemas (`stack` resource and `stack[s]` data sources) which provides:

- A boolean to indicate lock status
- Lock timestamp
- Who locked the stack
- The note they left (if any)

I can see this being useful in a number of scenarios. My particular use case is that I want to prevent changes to the stack when it's locked (not just the built-in "block run triggers" behavior). ~~I'm planning on doing this with a plan policy that reads this attribute and throws a warning or error if there are planned changes to the stack if it's locked.~~

To that effect, I've also added a `prevent_changes_when_locked` boolean input that will throw an error during planning if it's enabled, the stack is locked, and there are planned changes to the resource.

## Type of change

## Related issues

None

### Development

- [x] Lint rules pass locally
- [ ] ~~Examples for new resources and data sources have been added~~
- [ ] ~~Default values have been documented in the description (e.g., "Dummy: (Boolean) Blah blah. Defaults to `false`.)~~

### Code review

- [x] Changes have been reviewed by at least one other engineer

## Testing

Using a locally-build binary (built against this branch), you can see the `spacelift_stack`'s new attribute when the stack is locked and unlocked:

```console
$ tofu refresh && tofu console <<<'spacelift_stack.test1.lock'
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test]
tolist([
  {
    "locked" = false
    "locked_at" = 0
    "locked_by" = ""
    "note" = ""
  },
])

$ tofu refresh && tofu console <<<'spacelift_stack.test1.lock'
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test1]
tolist([
  {
    "locked" = true
    "locked_at" = 1789058682
    "locked_by" = "zach.himsel@addepar.com"
    "note" = "no touching!"
  },
])
```

### Testing the `prevent_changes_when_locked` control

<details><summary>unprotected, no changes</summary>

```console
$ tofu plan -var name="zhimsel-lock-test1" -var prevent_changes_when_locked="false"
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test]

No changes. Your infrastructure matches the configuration.
```

</details>

<details><summary>unprotected, changes</summary>

```console
$ tofu plan -var name="zhimsel-lock-test2" -var prevent_changes_when_locked="false"
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test]

OpenTofu will perform the following actions:

  # spacelift_stack.test1 will be updated in-place
  ~ resource "spacelift_stack" "test1" {
        id                               = "zhimsel-lock-test"
      ~ name                             = "zhimsel-lock-test1" -> "zhimsel-lock-test2"
        # (34 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

</details>

<details><summary>protected, no changes</summary>

When already protected:

```console
$ tofu plan -var name="zhimsel-lock-test1" -var prevent_changes_when_locked="true"
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test]

No changes. Your infrastructure matches the configuration.
```

When _enabling_ protection:

```console
$ tofu plan -var name="zhimsel-lock-test1" -var prevent_changes_when_locked="true"
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test]

OpenTofu will perform the following actions:

  # spacelift_stack.test1 will be updated in-place
  ~ resource "spacelift_stack" "test1" {
        id                               = "zhimsel-lock-test"
        name                             = "zhimsel-lock-test1"
      ~ prevent_changes_when_locked      = false -> true
        # (33 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

</details>

<details><summary>protected, changes, no lock note</summary>

```console
$ tofu plan -var name="zhimsel-lock-test2" -var prevent_changes_when_locked="true"
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test]

Planning failed. OpenTofu encountered an error while generating this plan.

╷
│ Error: stack is locked
│
│ This stack has `prevent_changes_when_locked` enabled, but it is currently locked
│   by: zach.himsel@addepar.com
│   at: 2026-09-10T16:48:26Z
│
│   with spacelift_stack.test1,
│   on stacks.tf line 30, in resource "spacelift_stack" "test1":
│   30: resource "spacelift_stack" "test1" {
╵
```

</details>

<details><summary>protected, changes, single-line lock note</summary>

```console
$ tofu plan -var name="zhimsel-lock-test2" -var prevent_changes_when_locked="true"
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test]

Planning failed. OpenTofu encountered an error while generating this plan.

╷
│ Error: stack is locked
│
│ This stack has `prevent_changes_when_locked` enabled, but it is currently locked
│   by: zach.himsel@addepar.com
│   at: 2026-09-10T16:49:11Z
│   with the note:
│
│ test note
│
│   with spacelift_stack.test1,
│   on stacks.tf line 30, in resource "spacelift_stack" "test1":
│   30: resource "spacelift_stack" "test1" {
╵
```

</details>

<details><summary>protected, changes, multi-line lock note</summary>

```console
$ tofu plan -var name="zhimsel-lock-test2" -var prevent_changes_when_locked="true"
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test]

Planning failed. OpenTofu encountered an error while generating this plan.

╷
│ Error: stack is locked
│
│ This stack has `prevent_changes_when_locked` enabled, but it is currently locked
│   by: zach.himsel@addepar.com
│   at: 2026-09-10T16:49:47Z
│   with the note:
│
│ test lock
│
│ with multiline message
│
│ how's this?
│
│   with spacelift_stack.test1,
│   on stacks.tf line 30, in resource "spacelift_stack" "test1":
│   30: resource "spacelift_stack" "test1" {
╵
```

</details>

<details><summary>already protected, replacement </summary>

```console
$ tofu apply -var name="zhimsel-lock-test1" -var prevent_changes_when_locked="true" -replace spacelift_stack.test1

OpenTofu will perform the following actions:

  # spacelift_stack.test1 will be replaced, as requested
-/+ resource "spacelift_stack" "test1" {
        # (16 unchanged attributes hidden)
    }

Plan: 1 to add, 0 to change, 1 to destroy.

spacelift_stack.test1: Destroying... [id=zhimsel-lock-test1]
╷
│ Error: stack is locked
│
│ This stack has `prevent_changes_when_locked` enabled, but it is currently locked
│   by: zach.himsel@addepar.com
│   at: 2026-09-10T16:52:10Z
╵
```

</details>

<details><summary>already protected, destroy </summary>

```console
$ tofu destroy -var name="zhimsel-lock-test1" -var prevent_changes_when_locked="true"

OpenTofu will perform the following actions:

  # spacelift_stack.test1 will be destroyed
  - resource "spacelift_stack" "test1" {
        # (16 unchanged attributes hidden)
    }

Plan: 0 to add, 0 to change, 1 to destroy.

spacelift_stack.test1: Destroying... [id=zhimsel-lock-test1]
╷
│ Error: stack is locked
│
│ This stack has `prevent_changes_when_locked` enabled, but it is currently locked
│   by: zach.himsel@addepar.com
│   at: 2026-09-10T17:01:57Z
╵
```


zhimsel:
> I'm planning on doing this with a plan policy that reads this attribute and throws a warning or error if there are planned changes to the stack if it's locked.

Actually, we may be able to do that within the provider itself. Gonna hack on it.

EDIT: yep, I got something worked out. Just pushed a new commit to include a boolean input that produces an error (when `true`) if the resource includes changes to the stack. I'll update the description above and include additional test results.

docs/resources/stack.md
```diff
@@ -261,6 +261,7 @@ resource "spacelift_role_attachment" "spacelift-admin-operator" {
 - `labels` (Set of String)
 - `manage_state` (Boolean) Determines if Spacelift should manage state for this stack. Defaults to `true`.
 - `opentofu` (Block List, Max: 1) OpenTofu-specific configuration. Presence means this Stack is a native OpenTofu Stack. (see [below for nested schema](#nestedblock--opentofu))
+- `prevent_changes_when_locked` (Boolean) If true, Terraform will fail during planning when the stack is locked and the plan includes changes to this resource.
```
zhimsel: I'm open to suggestions for alternate names for this parameter. I also considered `protect_when_locked` (but thought that may suggest it's something being enabled in Spacelift itself, since that's what `protect_from_deletion` does, but maybe that's just me 😄). Happy to change it to whatever y'all would prefer.

docs/resources/stack.md
```diff
@@ -261,6 +261,7 @@ resource "spacelift_role_attachment" "spacelift-admin-operator" {
 - `labels` (Set of String)
 - `manage_state` (Boolean) Determines if Spacelift should manage state for this stack. Defaults to `true`.
 - `opentofu` (Block List, Max: 1) OpenTofu-specific configuration. Presence means this Stack is a native OpenTofu Stack. (see [below for nested schema](#nestedblock--opentofu))
+- `prevent_changes_when_locked` (Boolean) If true, Terraform will fail during planning when the stack is locked and the plan includes changes to this resource. Defaults to `false`.
```
zhimsel, line 264: > Defaults to `false`.

Right now, the property in state defaults to `null` if it's not explicitly set (but it behaves like `false`). Should we make that default to `false` in the state?

zhimsel: Never mind, this is actually a good thing. That means when upgrading provider versions, and using the default value, there will be a no-op diff.

spacelift/resource_stack.go
```diff
@@ -54,6 +55,40 @@ func resourceStack() *schema.Resource {
 				return nil
 			}
 
+			if prevent, ok := diff.GetOk("prevent_changes_when_locked"); ok && prevent.(bool) {
```
rekfuki, line 58: This only checks the planned value, so you can flip the flag off and change other stuff in the same plan. Probably want the prior value too, same as `how protect_from_deletion behaves`:

```suggestion
			oldPrevent, newPrevent := diff.GetChange("prevent_changes_when_locked")
			if oldPrevent.(bool) || newPrevent.(bool) {
```

spacelift/resource_stack.go
```diff
@@ -631,6 +695,12 @@ func resourceStack() *schema.Resource {
 				Optional:    true,
 				Description: "Project globs is an optional list of paths to track changes of in addition to the project root.",
 			},
+			"prevent_changes_when_locked": {
+				Type:        schema.TypeBool,
+				Description: "If true, Terraform will fail during planning when the stack is locked and the plan includes changes to this resource.",
```
rekfuki: Maybe "in-place changes", with a note that destroy/replace aren't covered? CustomizeDiff doesn't run on destroy/removal, and `-replace` comes through with an empty diff.
 
Also, "Defaults to `false`." needs to go in the Go description, not the .md, or `go generate` will strip it and the docs check fails.

spacelift/resource_stack_test.go
```diff
@@ -110,11 +116,12 @@ func TestStackResource(t *testing.T) {
 				ImportStateVerifyIgnore: []string{"import_state"},
```
rekfuki: Read doesn't set this, so import verify will fail (same as `import_state`). Same for the other import steps.

```suggestion
				ImportStateVerifyIgnore: []string{"import_state", "prevent_changes_when_locked"},
```

Or set it in `PopulateStack` with `d.Set(..., d.Get(...))`, which also avoids the diff from null to false after import.

spacelift/internal/structs/stack.go
```diff
@@ -290,6 +293,23 @@ func PopulateStack(d *schema.ResourceData, stack *Stack) diag.Diagnostics {
 	d.Set("space_id", stack.Space)
 	d.Set("slug", stack.ID)
 
+	lockBlock := map[string]any{
```
rekfuki: nit: this is duplicated in `data_stack.go` and `data_stacks.go`. We usually do a method on `*Stack` for this, see `IaCSettings()` / `VCSSettings()`.
 
Also non-blocking: an empty `lock` list when unlocked would match the other nested blocks and you could drop `locked`.

rekfuki (changes requested):
Thanks, this looks good overall. Heads up that it needed a backend change (lock fields weren't exposed to machine sessions), that's in now.

A few comments:

spacelift/resource_stack.go
zhimsel: I'd actually consider that a feature, not a bug.

Also, critically: `prevent_changes_when_locked` is different from `protect_from_deletion`. The latter is an actual stack setting _in Spacelift itself_. The former is a "Terraform directive" that only exists in state and code. So it doesn't actually require an _apply_ to take effect. It's closer in concept/behavior to the Terraform builtin `lifecycle { prevent_destroy }`.

spacelift/resource_stack.go
zhimsel: > Maybe "in-place changes", with a note that destroy/replace aren't covered? CustomizeDiff doesn't run on destroy/removal, and `-replace` comes through with an empty diff.

That's a good point, and should be documented. Will update.

> Also, "Defaults to `false`." needs to go in the Go description, not the .md, or `go generate` will strip it and the docs check fails.

👍🏻 will do

zhimsel: Hm, question: _should_ we have this fail if a stack is locked and is going to be destroyed/replaced? Putting myself in the shoes of a user (who didn't write this PR), I would _expect_ that to be the behavior, I think.

EDIT: I think we should.

spacelift/resource_stack_test.go
zhimsel: > Or set it in `PopulateStack` with `d.Set(..., d.Get(...))`, which also avoids the diff from null to false after import.

I opted to [go with your first suggestion](https://github.com/spacelift-io/terraform-provider-spacelift/pull/846/commits/4a164f8b5514fae52677f43ad6f99398ba476732
), primarily because `PopulateStack` sets the state from _API data_. Since this property isn't in the Spacelift API itself (its an internal provider flag), that seems like it may cause some confusion down the road.

spacelift/internal/structs/stack.go
zhimsel: > nit: this is duplicated in `data_stack.go` and `data_stacks.go`. We usually do a method on `*Stack` for this, see `IaCSettings()` / `VCSSettings()`.

Done [here](https://github.com/spacelift-io/terraform-provider-spacelift/pull/846/commits/93fc9909780efff4a89dc17c00373fd4d2d6d986).

zhimsel: > Also non-blocking: an empty `lock` list when unlocked would match the other nested blocks and you could drop `locked`.

Hm, I had a bit of a think about this, and I think I'd argue for keeping it an always-present (non-empty) map (with a boolean `locked`). This is for a few reasons. 

One: the other nested blocks are different in a key way: all the others are (writable) _configurations_, whereas this is a (read-only) _status_. So it's an apples-to-oranges comparison.

Two: having the keys always be available in the resource attributes makes downstream consumption much easier (and this being a read-only attribute, this is much more likely than the other blocks). For example: I originally planned on using the lock status in a plan policy that would prevent or warn against changes (I still might, even if `prevent_changes_when_locked` is enabled); having `lock.locked` always be present simplifies the Rego code a bit instead of needing to do "does `lock.locked` exist?". The same can be said for downstream Terraform consumers.

zhimsel:
> Heads up that it needed a backend change (lock fields weren't exposed to machine sessions), that's in now.

Ah, interesting. I didn't notice that in my testing since I was using a local `tofu` session with personal credentials (via `spacectl profile export-token`), not a machine API token (or stack permissions). I would've _eventually_ run into that, but you caught it way quicker than I. Thanks 😄

zhimsel:
I re-ran all the practical tests with the new changes, and added some more tests. I updated the PR description with the test changes.

spacelift/resource_stack.go
rekfuki: Fair enough, the `prevent_destroy` comparison makes sense.

I think the apply-time check might not agree with it though, see the comment on `resourceStackUpdate`.

spacelift/resource_stack.go
```diff
@@ -697,7 +650,7 @@ func resourceStack() *schema.Resource {
 			},
 			"prevent_changes_when_locked": {
 				Type:        schema.TypeBool,
-				Description: "If true, Terraform will fail during planning when the stack is locked and the plan includes changes to this resource.",
```
rekfuki, line 700: Looks good. One small thing: the destroy block runs in delete, so destroy plans clean and then fails on apply, same as protect_from_deletion. Might be worth a few words in the description so nobody reads "will fail" as plan-time for destroy.

spacelift/resource_stack.go
```diff
@@ -1006,6 +1019,10 @@ func resourceStackRead(ctx context.Context, d *schema.ResourceData, meta any) di
 }
 
 func resourceStackUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
+	if diags := checkStackLock(d); diags.HasError() {
+		return diags
+	}
+
 	var ret diag.Diagnostics
```
rekfuki, line 1026: I think this ends up disagreeing with the plan-time check. If I read it right, `CustomizeDiff` looks at the planned flag and skips flag-only changes, so a plan that sets `prevent_changes_when_locked = false` on a locked stack should go through. Then this falls back to prior state, sees `true`, and fails the apply. So plan passes, apply fails, and while the stack is locked I don't see an apply that can turn the flag off. Could you double check that scenario?

Also not sure the update check buys us anything because`CustomizeDiff` already covers in-place changes at plan time, and as far as I can tell `lock.0.locked` here is the same plan-time value, so it wouldn't catch a lock taken after the plan either. If that's right, maybe just drop it from Update and keep it in delete:

```suggestion
func resourceStackUpdate(ctx context.Context, d *schema.ResourceData, meta any) diag.Diagnostics {
	var ret diag.Diagnostics
```

That would also give a way out for a locked stack removed from config: add it back with the flag off, apply, destroy.

spacelift/resource_stack.go
rekfuki: Agreed, if the flag says "prevent changes" then destroy counts as a change.

One small thing: the destroy block runs in delete, so tofu destroy plans clean and then fails on apply, same as `protect_from_deletion`. Might be worth a few words in the description so nobody reads "will fail" as plan-time for destroy.

spacelift/internal/structs/stack.go
rekfuki: Makes sense, status vs config is a fair distinction

rekfuki (changes requested):
Thanks for the quick turnaround, most of it looks good. One thing I'm not sure about with the new apply-time check

Edit: Grrr I see that the review comments were duplicated. It appears that my replies are now both in the original conversations as well as on their own. Sorry about that.

zhimsel:
> Edit: Grrr I see that the review comments were duplicated. It appears that my replies are now both in the original conversations as well as on their own. Sorry about that.

Yeah, that's one of those annoying Github behaviors. When comments are replies to existing threads, and made as part of a review, it displays them twice. All good, I'm used to it 😄

spacelift/resource_stack.go
zhimsel: Yeah, the "destroy only fails during apply, not planning" thing is unfortunate. It's a side-effect of the particular framework version this provider uses. As far as I can tell, and as far as my testing has revealed, there's no way to trigger the destroy failure during planning. It has to do with how the HCL is evaluated during each of those phases, if I understand it correctly.

So... not as bad as _not failing at all_, but not as good as failing during plan.

Long-term, from what I understand, once this resource is migrated away from the SDKv2 (to the plugin framework), the error _can_ be moved to planning. To that effect, I [added a TODO comment](https://github.com/spacelift-io/terraform-provider-spacelift/pull/846/commits/c9485b458ab3cde930c9c505425b1986c6bcc927) to remind whomever does the migration to account for that.

zhimsel: > Might be worth a few words in the description so nobody reads "will fail" as plan-time for destroy.

Agreed. I'll push up a description change.

spacelift/resource_stack.go
zhimsel: [Done](https://github.com/spacelift-io/terraform-provider-spacelift/pull/846/commits/c1f811f8300cc830db2553685c21bc1817bebd08)

spacelift/resource_stack.go
zhimsel: > So plan passes, apply fails, and while the stack is locked I don't see an apply that can turn the flag off. Could you double check that scenario?

Hm, very interesting. You may be right. I'll get a test scenario set up to validate. I should probably also check the difference between running a `destroy` command (or using `-replace`/tainting) vs _removing_ the HCL from the config. Not sure if the SDK differentiates those. Theoretically in the former, it _should_ have access to the HCL config.

zhimsel: Your deadlock scenario is confirmed. Trying to turn off the safeguard for a stack that's locked:

```console
$ tofu apply -var name="zhimsel-lock-test1" -var prevent_changes_when_locked="false"
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test1]
spacelift_environment_variable.this: Refreshing state... [id=stack/zhimsel-lock-test1/foo]

OpenTofu used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place (current -> planned)

OpenTofu will perform the following actions:

  # spacelift_stack.test1 will be updated in-place
  ~ resource "spacelift_stack" "test1" {
        id                               = "zhimsel-lock-test1"
        name                             = "zhimsel-lock-test1"
      ~ prevent_changes_when_locked      = true -> false
        # (33 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  OpenTofu will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

spacelift_stack.test1: Modifying... [id=zhimsel-lock-test1]
╷
│ Error: stack is locked
│
│   with spacelift_stack.test1,
│   on stacks.tf line 30, in resource "spacelift_stack" "test1":
│   30: resource "spacelift_stack" "test1" {
│
│ This stack has `prevent_changes_when_locked` enabled, but it is currently locked
│   by: zach.himsel@addepar.com
│   at: 2026-09-14T17:20:34Z
```

Plan works, but apply fails.

EDIT: it appears that the state is actually updated; so subsequent plans/applies succeed, with the setting correctly flipped. But that _initial_ apply will fail from the user's perspective.

zhimsel: And after [removing the check from `resourceStackUpdate`](https://github.com/spacelift-io/terraform-provider-spacelift/pull/846/commits/d7929648c36e95a55b33ea553a622a66cd801cff): 

```console
$ tofu apply -var name="zhimsel-lock-test1" -var prevent_changes_when_locked="false"
spacelift_stack.test1: Refreshing state... [id=zhimsel-lock-test1]
spacelift_environment_variable.this: Refreshing state... [id=stack/zhimsel-lock-test1/foo]

OpenTofu used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  ~ update in-place (current -> planned)

OpenTofu will perform the following actions:

  # spacelift_stack.test1 will be updated in-place
  ~ resource "spacelift_stack" "test1" {
        id                               = "zhimsel-lock-test1"
        name                             = "zhimsel-lock-test1"
      ~ prevent_changes_when_locked      = true -> false
        # (33 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  OpenTofu will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

spacelift_stack.test1: Modifying... [id=zhimsel-lock-test1]
spacelift_stack.test1: Modifications complete after 0s [id=zhimsel-lock-test1]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.

```

zhimsel: Nice catch 😄

zhimsel: I think the only remaining scenario where the protection field is "stuck" is:

- Stack is currently locked
- `prevent_changes_when_locked` is `true` in state (i.e. the previous apply had it enabled)
- Stack is trying to be deleted

There's really only two ways around this... either:

- _unlock the stack_ (seems like the simplest option 😄)
or
- disable the protection, apply, then destroy

There's just not really a good way around that. TF doesn't evaluate the HCL during a destroy; it can only go off the last state of a resource.

rekfuki: Thanks for taking the time to actually run through it, appreciated.

Agreed, that's fine. It mirrors `protect_from_deletion`, where you also have to flip the flag and apply before a destroy goes through. And unlocking is the whole point of the feature anyway, so having it as the way out makes sense.

rekfuki (dismissed):
Thanks for sticking with this and for testing every scenario along the way 

LGTM 🎉 ✅

zhimsel:
Thanks for reviewing! Looking forward to release 😄

rekfuki (approved):

rekfuki:
Merged - appreciate the submission 🎉 Will try to get it released promptly

PR #852, opened 2026-09-14
fix(stack): retain the assigned worker pool when it is not set in configuration

sephriot:
## Description of the change

A stack managed without an explicit worker pool no longer has its pool reset on every update — it keeps the pool it currently has. This removes the update failure customers hit when drift detection was enabled and the pool had been attached outside their configuration.

Worth a reviewer's decision: this also means dropping the attribute from a configuration no longer moves a stack back to the public pool. That is the same trade-off this resource already makes for the space, slug and version attributes.

## Type of change

## Related issues

> https://app.clickup.com/t/869e2gx9d

### Development

- [x] Lint rules pass locally

### Code review


dabit2002:
Task linked: [CU-869e2gx9d Improve error message for drift detection worker pool configuration conflict](https://app.clickup.com/t/869e2gx9d)

sephriot:
Closing: the reset-on-every-update behaviour is not the right thing to change here. Handling this on the backend instead.

